PR #6, opened 2020-05-22
Fix minor bugs

soekarnohatta:
This fixes:
https://github.com/PaulSonOfLars/gotgbot/blob/5c2fca1805f1963e05f500c0fd9ade414a3eefbf/ext/sendables.go#L903
https://github.com/PaulSonOfLars/gotgbot/blob/5c2fca1805f1963e05f500c0fd9ade414a3eefbf/ext/chat_sendables.go#L281
https://github.com/PaulSonOfLars/gotgbot/blob/5c2fca1805f1963e05f500c0fd9ade414a3eefbf/ext/User.go#L22

### DIFF
--- a/ext/User.go
+++ b/ext/User.go
@@ -20,5 +20,10 @@ type UserProfilePhotos struct {
 
 // GetProfilePhotos Retrieves a user's profile pictures
 func (user User) GetProfilePhotos() (*UserProfilePhotos, error) {
-	return user.Bot.GetUserProfilePhotos(user.Id)
+	return user.Bot.GetUserProfilePhotos(user.Id, 0, 0)
+}
+
+// GetProfilePhotosOffset Retrieves a user's profile pictures
+func (user User) GetProfilePhotosOffset(offset int, limit int) (*UserProfilePhotos, error) {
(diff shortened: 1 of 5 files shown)


ext/User.go
```diff
@@ -19,6 +19,6 @@ type UserProfilePhotos struct {
 }
 
 // GetProfilePhotos Retrieves a user's profile pictures
-func (user User) GetProfilePhotos(offset int, limit int) (*UserProfilePhotos, error) {
```
PaulSonOfLars, line 22: https://core.telegram.org/bots/api#getuserprofilephotos
If youre removing the offsets for the base image, might be worth adding a `GetProfilePhotosOffset(offset int, limit int) which allows you to iterate, since tg exposes that info.

soekarnohatta: > https://core.telegram.org/bots/api#getuserprofilephotos
> If you're removing the offsets for the base function, might be worth adding a `GetProfilePhotosOffset(offset int, limit int)` which allows you to iterate, since tg exposes that info.

I'll check tomorrow morning

soekarnohatta: please check newes commit

PR #7, opened 2020-05-24
Fix trailing slash issue on Webhooks URL

dhruvkelawala:


updater.go
```diff
@@ -188,7 +188,12 @@ func (u Updater) SetWebhook(path string, webhook Webhook) (bool, error) {
 	}
 
 	v := url.Values{}
-	v.Add("url", webhook.URL+"/"+path)
+	if webhook.URL[len(webhook.URL)-1:] != "/" {
```
PaulSonOfLars, line 191: Recommend using `strings.HasSuffix()` instead of this

updater.go
```diff
@@ -188,7 +188,12 @@ func (u Updater) SetWebhook(path string, webhook Webhook) (bool, error) {
...
+		v.Add("url", webhook.URL+"/"+path)
```
PaulSonOfLars, line 192: The `v.Add()` is irrelevant to this if statement. Could you simplify this to a variable, which takes the default, such as
```
webhookURL:= webhook.URL+path
if webhook.URL has suffix...
    Remove the /

v.add(webhookURL)
```

Clearly shows the default case and that having a trailing / is an edge case, removes an unnecessary else statement, and only does one add call with a clearly named variable

updater.go
dhruvkelawala: Got it.

PR #14, opened 2021-01-23
Added regex filter

itzKassem:


ext/handlers/filters/messages.go
```diff
@@ -32,6 +33,19 @@ func Text(msg *gotgbot.Message) bool {
 	return msg.Text != ""
 }
 
+func Regex(text string) Message {
+	return func(msg *gotgbot.Message) bool {
+		   if msg.Text == "" {
```
PaulSonOfLars, line 38: nit: Please fix the indentation on these statements. Golang indents are tabs, not spaces

PaulSonOfLars, line 38: Should the regex filter also work on captions?

ext/handlers/filters/messages.go
```diff
@@ -32,6 +33,19 @@ func Text(msg *gotgbot.Message) bool {
...
+				return false
+			}
+           res, err:= regexp.Match(text, []byte(msg.Text))
```
PaulSonOfLars, line 41: You can improve performance by calling `r := regexp.Compile()` in the function body, and using `r.Match()` here instead.
This will avoid having to compile the same regex on every update. It will also allow you to return any errors raised in the `Compile`, which are currently swallowed with no details (making debugging difficult).

Also, I recommend using `MatchString` here, which will avoid the memory copy caused by the type conversion.

ext/handlers/filters/messages.go
```diff
@@ -32,6 +33,19 @@ func Text(msg *gotgbot.Message) bool {
...
+	       if err != nil {
```
PaulSonOfLars, line 42: nit: indents again

PaulSonOfLars (changes requested):
Thank you for your contribution!

A good start, some minor comments on style/error handling

Also, as a sidenote: Would be nice to see a PR description explaining the "what" of the PR, so that one doesnt need to open the code :)

ext/handlers/filters/messages.go
```diff
@@ -2,6 +2,7 @@ package filters
 
 import (
 	"github.com/PaulSonOfLars/gotgbot/v2"
+        "regexp"
```
PaulSonOfLars, line 5:
```suggestion
	"regexp"

	"github.com/PaulSonOfLars/gotgbot/v2"
```

Usual import order is:
- stdlib
- external packages
- local package
with a newline between each category, and each is alphabetically ordered

I appreciate that this isnt clear right now, I'll add some linters to the repo to make sure these are all automatically covered soon.

PaulSonOfLars:
Closing, since this has been implemented in #27

ext/handlers/filters/messages.go
itzKassem: i was completely new to golang :)

itzKassem: why not

PR #18, opened 2021-03-29
Fixed invalid link

wemiprog:
A link in the readme was only github.com without https://, what made him relative and so not working. Only added this = 1 line, 8 letters.
So, shouln't have any (negative) impact


PaulSonOfLars (approved):
Thank you for catching and fixing!

PR #20, opened 2021-05-05
handlers: implement generic update handler

JaskaranSM:
# What

Describe the intent of your changes in detail, and what they are resolving. 
    Adds a new handler, which basically lets all the Updates pass through to provided Response functions.

Provide links to external sources if relevant (eg: github issues, external discussions, etc)

# Impact

- Have you included documentation, or updated existing documentation? No

### DIFF
--- a/ext/handlers/command.go
+++ b/ext/handlers/command.go
@@ -27,6 +27,9 @@ func NewCommand(c string, r Response) Command {
 }
 
 func (c Command) CheckUpdate(b *gotgbot.Bot, u *gotgbot.Update) bool {
+	if u.Message == nil {
+		return false
+	}
 	if u.Message != nil {
 		if u.Message.Text == "" && u.Message.Caption == "" {
 			return false
--- a/ext/handlers/message.go
+++ b/ext/handlers/message.go
@@ -25,6 +25,9 @@ func NewMessage(f filters.Message, r Response) Message {
 }
 
 func (m Message) CheckUpdate(b *gotgbot.Bot, u *gotgbot.Update) bool {
+	if u.Message == nil {
+		return false
+	}
 	if u.Message != nil {
 		return m.Filter == nil || m.Filter(u.Message)
 	}
--- a/ext/handlers/update.go
+++ b/ext/handlers/update.go
@@ -1,0 +1,30 @@
+package handlers
+
+import (
+	"fmt"
+
+	"github.com/PaulSonOfLars/gotgbot/v2"
+	"github.com/PaulSonOfLars/gotgbot/v2/ext"
+)
+
+type Update struct {
+	Response Response
+}
+
+func NewUpdate(r Response) Update {
+	return Update{
+		Response: r,
+	}
+}
+
+func (u Update) CheckUpdate(b *gotgbot.Bot, upd *gotgbot.Update) bool {
+	return upd != nil
+}
+
+func (u Update) HandleUpdate(b *gotgbot.Bot, ctx *ext.Context) error {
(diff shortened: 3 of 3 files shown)


ext/handlers/command.go
```diff
@@ -27,6 +27,9 @@ func NewCommand(c string, r Response) Command {
 }
 
 func (c Command) CheckUpdate(b *gotgbot.Bot, u *gotgbot.Update) bool {
+	if u.Message == nil {
```
PaulSonOfLars, line 30: This breaks command handling on edited and channel messages, which have nil `Messages` structs. See later in this function for how these are handled.

ext/handlers/message.go
```diff
@@ -25,6 +25,9 @@ func NewMessage(f filters.Message, r Response) Message {
 }
 
 func (m Message) CheckUpdate(b *gotgbot.Bot, u *gotgbot.Update) bool {
+	if u.Message == nil {
```
PaulSonOfLars, line 28: same comment as for command handlers; this breaks channels and edited messages.

ext/handlers/update.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func NewUpdate(r Response) Update {
+	return Update{
+		Response: r,
+	}
+}
+
+func (u Update) CheckUpdate(b *gotgbot.Bot, upd *gotgbot.Update) bool {
+	return upd != nil
```
PaulSonOfLars, line 21: Not having any sort of filters on this handlers makes it a firehose of data receiving every single update from the API, where most of them aren't going to be relevant. This defeats the current definition of handlers, which were chosen to reduce the number updates needing processing, which avoids running on all of them.

Taking this into account, I see two ways forward fir this PR:
1. Define a handler (+filters) for each update type, and specifically DON'T create the raw `Update` handler. This helps keep the library clear, and will discourage the antipattern of creating one `Update` handler to do everything at the same time.
1. Keep this `Update` handler, but make it more configurable to reduce unwanted data. For example, it would be nice to have a way to create filters on updates, like the other handlers.
    - This would also allow us to define other handlers in terms of `Update` handlers. I dont know how feasible this is, or if its a good idea - but could be nice to reduce the duplication.

The idea is that if you give users a raw `Update` handler, they are more likely to end up getting nil pointers from trying to access the wrong update type. This would be an antipattern, since:
1. It would encourage people to create one handler ~~to rule them all~~  to run everything, which could lead to confusing code.
1. It makes code much more error prone, since nil pointers would be more likely.
1. Those errors can't be detected at build time (It only breaks at runtime when faced with an unexpected update)
1. The fix to those errors is to add a check and define `u.Message != nil` etc, at which point the code might as well be using a MessageHandler (or other relevant handler)

What do you think?

PaulSonOfLars (commented):
Not a fan of the Idea of a global `Update` handler with no filtering capabilities, since I believe it would promote using the library in an antipattern fashion. See comment for more details. 

Also, two small nits where the added checks actually break existing features.

Otherwise, code looks good!

ext/handlers/update.go
JaskaranSM: IMO the first way you mentioned is better. It is very likely that user wanna handle multiple update types. So creating update handlers for the update types that are accepted in allowed_updates is more feasible here.

PaulSonOfLars: If you'd like to do that, I'd be happy to review a follow up. Doesn't need to be all the handlers, can be just the ones you need for now :)

JaskaranSM:
Closing this, will PR individual handlers.

PR #21, opened 2021-05-08
handlers: implement chatmember update handler

JaskaranSM:
# What

Describe the intent of your changes in detail, and what they are resolving. 
Added new ChatMember update handler

Provide links to external sources if relevant (eg: github issues, external discussions, etc)

# Impact

- Are your changes backwards compatible? yes
- Have you included documentation, or updated existing documentation? no 

### DIFF
--- a/ext/handlers/chatmember.go
+++ b/ext/handlers/chatmember.go
@@ -21,13 +21,7 @@ func NewChatMember(f filters.ChatMember, r Response) ChatMember {
 }
 
 func (c ChatMember) CheckUpdate(b *gotgbot.Bot, u *gotgbot.Update) bool {
-	if u.ChatMember != nil {
-		if c.Filter != nil {
-			return c.Filter(u.ChatMember)
-		}
-		return true
-	}
-	return false
+	return u.ChatMember != nil && c.Filter != nil && c.Filter(u.ChatMember)
 }
 
 func (c ChatMember) HandleUpdate(b *gotgbot.Bot, ctx *ext.Context) error {
--- a/ext/handlers/chatmember.go
+++ b/ext/handlers/chatmember.go
@@ -1,0 +1,39 @@
+package handlers
+
+import (
+	"fmt"
+
+	"github.com/PaulSonOfLars/gotgbot/v2"
+	"github.com/PaulSonOfLars/gotgbot/v2/ext"
+	"github.com/PaulSonOfLars/gotgbot/v2/ext/handlers/filters"
+)
+
+type ChatMember struct {
+	Response Response
+	Filter   filters.ChatMember
+}
+
+func NewChatMember(f filters.ChatMember, r Response) ChatMember {
+	return ChatMember{
+		Response: r,
+		Filter:   f,
+	}
+}
+
+func (c ChatMember) CheckUpdate(b *gotgbot.Bot, u *gotgbot.Update) bool {
+	if u.ChatMember != nil {
+		if c.Filter != nil {
+			return c.Filter(u.ChatMember)
+		}
+		return true
+	}
+	return false
+}
+
+func (c ChatMember) HandleUpdate(b *gotgbot.Bot, ctx *ext.Context) error {
+	return c.Response(b, ctx)
+}
+
+func (c ChatMember) Name() string {
+	return fmt.Sprintf("chatmember_%p", c.Response)
+}
--- a/ext/handlers/chatmember.go
+++ b/ext/handlers/chatmember.go
@@ -21,7 +21,11 @@ func NewChatMember(f filters.ChatMember, r Response) ChatMember {
 }
 
 func (c ChatMember) CheckUpdate(b *gotgbot.Bot, u *gotgbot.Update) bool {
-	return u.ChatMember != nil && c.Filter != nil && c.Filter(u.ChatMember)
+	if u.ChatMember == nil {
+		return false
+	}
+	
+	return c.Filter == nil || c.Filter(u.ChatMember)
 }
 
 func (c ChatMember) HandleUpdate(b *gotgbot.Bot, ctx *ext.Context) error {
--- a/ext/handlers/filters/updates.go
+++ b/ext/handlers/filters/updates.go
@@ -1,0 +1,61 @@
+package filters
+
+import (
+	"github.com/PaulSonOfLars/gotgbot/v2"
+)
+
+type ChatMember func(u *gotgbot.ChatMemberUpdated) bool
+
+func (f ChatMember) And(f2 ChatMember) ChatMember {
+	return func(cq *gotgbot.ChatMemberUpdated) bool {
+		return f(cq) && f2(cq)
+	}
+}
+
+func (f ChatMember) Or(f2 ChatMember) ChatMember {
+	return func(cq *gotgbot.ChatMemberUpdated) bool {
+		return f(cq) || f2(cq)
+	}
+}
+
+func (f ChatMember) Not() ChatMember {
+	return func(cq *gotgbot.ChatMemberUpdated) bool {
+		return !f(cq)
+	}
+}
+
+func ChatMemberUserId(id int64) ChatMember {
+	return func(cm *gotgbot.ChatMemberUpdated) bool {
+		return cm.NewChatMember.User.Id == id
+	}
+}
+
+func PerformerUserId(id int64) ChatMember {
+	return func(cm *gotgbot.ChatMemberUpdated) bool {
+		return cm.From.Id == id
+	}
+}
+
+func ChatMemberChatId(id int64) ChatMember {
+	return func(cm *gotgbot.ChatMemberUpdated) bool {
+		return cm.Chat.Id == id
+	}
+}
+
+func ChatMemberAll() ChatMember {
+	return func(_ *gotgbot.ChatMemberUpdated) bool {
+		return true
+	}
+}
+
+func ChatMemberGroup() ChatMember {
+	return func(cm *gotgbot.ChatMemberUpdated) bool {
+		return cm.Chat.Type == "group"
+	}
+}
+
+func ChatMemberSupergroup() ChatMember {
+	return func(cm *gotgbot.ChatMemberUpdated) bool {
+		return cm.Chat.Type == "supergroup"
+	}
+}
--- a/ext/handlers/filters/updates.go
+++ b/ext/handlers/filters/updates.go
@@ -59,3 +59,9 @@ func ChatMemberSupergroup() ChatMember {
 		return cm.Chat.Type == "supergroup"
 	}
 }
+
+func ChatMemberChannel() ChatMember {
+	return func(cm *gotgbot.ChatMemberUpdated) bool {
+		return cm.Chat.Type == "channel"
+	}
+}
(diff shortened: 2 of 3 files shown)


ext/handlers/chatmember.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func (c ChatMember) CheckUpdate(b *gotgbot.Bot, u *gotgbot.Update) bool {
+	if u.ChatMember != nil {
+		if c.Filter != nil {
+			return c.Filter(u.ChatMember)
+		}
+		return true
+	}
+	return false
```
PaulSonOfLars:
```suggestion
	return u.ChatMember != nil && c.Filter != nil && c.Filter(u.ChatMember)
```
This can be simplified into a single line; reducing unnecessary if statements improves readability

ext/handlers/filters/updates.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+}
+
+func ChatMemberUserId(id int64) ChatMember {
+	return func(cm *gotgbot.ChatMemberUpdated) bool {
+		return cm.NewChatMember.User.Id == id
+	}
+}
+
+func PerformerUserId(id int64) ChatMember {
```
PaulSonOfLars: "performer" is a new term that hasn't been used in other contexts. What do you think of changing this to `ChatMemberFromUserId`?

ext/handlers/filters/updates.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+}
+
+func ChatMemberAll() ChatMember {
+	return func(_ *gotgbot.ChatMemberUpdated) bool {
+		return true
+	}
+}
+
+func ChatMemberGroup() ChatMember {
```
PaulSonOfLars, line 51: Can we get a channel type here too? Could be useful for determining channel joins.

PaulSonOfLars (commented):
Thank you for your contribution! This is a great addition to the library.

Got a few comments, let me know what you think.

Also, this PR seems like a good place to add the MyChatMember handler at the same time - does that sound ok? Should be the same filters since they act in the same struct type, so only needs handler definition :)

ext/handlers/chatmember.go
JaskaranSM: i did that cuz i considered what if a user provides nil for the filter in constructor.

ext/handlers/filters/updates.go
JaskaranSM: Ok

ext/handlers/filters/updates.go
JaskaranSM: Ok and message filters doesnt have channel too.

ext/handlers/chatmember.go
PaulSonOfLars: The filter check is fine, the issue is the unnecessary double if statement which can be simplified

ext/handlers/filters/updates.go
PaulSonOfLars: Feel free to add it then!

ext/handlers/filters/updates.go
JaskaranSM: Done.

ext/handlers/filters/updates.go
JaskaranSM: > Can we get a channel type here too? Could be useful for determining channel joins.

Done

ext/handlers/chatmember.go
```diff
@@ -0,0 +1,33 @@
...
+
+func NewChatMember(f filters.ChatMember, r Response) ChatMember {
+	return ChatMember{
+		Response: r,
+		Filter:   f,
+	}
+}
+
+func (c ChatMember) CheckUpdate(b *gotgbot.Bot, u *gotgbot.Update) bool {
+	return u.ChatMember != nil && c.Filter != nil && c.Filter(u.ChatMember)
```
PaulSonOfLars: Just a question; opinions are welcome.

If Filter is nil, then this handler is completely pointless, and will always return false. That could lead to unexpected, hard to debug behaviour.

So the question is; would we rather:
a) have this "error" be swallowed up and have no indication that there is a bad thing happening (ie, the `Filter != nil` check is present, like in this PR)
b) fail with a runtime error because we're trying to call a nil function (potentially bad if happens in prod; would hopefully be tested first)

ext/handlers/filters/updates.go
```diff
@@ -0,0 +1,67 @@
+package filters
+
+import (
+	"github.com/PaulSonOfLars/gotgbot/v2"
+)
+
+type ChatMember func(u *gotgbot.ChatMemberUpdated) bool
+
+func (f ChatMember) And(f2 ChatMember) ChatMember {
```
PaulSonOfLars, line 9: Looking for feedback, not changes for this PR; do you consider the And/Or/Not method to be necessary/helpful? Is it not easier to create a filter chain with a custom func instead? They seem confusing.

PaulSonOfLars (approved):
Looks good to me! Got two small questions that I'd like your opinions on before in merge, but otherwise ready to go :) 

Thank you for the contribution!

ext/handlers/filters/updates.go
JaskaranSM: imo it is good, however considering go's on the fly anonymous functions, it can be done that way too. But from the user's perspective, i think it is nice to have a "helper" thingy for it.

ext/handlers/chatmember.go
JaskaranSM: I would strongly advice against the code that can lead to potential runtime errors. One should never de-reference a nil pointer. we should make the filter callback mandatory.

ext/handlers/filters/updates.go
PaulSonOfLars: Is it though? Or does it just add more confusion around how to pair filters? 

Personally, i think it would be much easier to read a filter using anonymous functions (which will use golang operators) rather than these, which have dubious operator precedence

ext/handlers/chatmember.go
PaulSonOfLars: I definitely agree - the question is, how do we define a nil filter?

In the current case, a nil filter == always false, handler never gets triggered.

However, i would argue that if I'm adding a new handler, and I set a nil filter, I would expect that handler to catch EVERY update; since I am not filtering on it. Does that make sense?

JaskaranSM: Thats why i said the filter must be there. Lets say we have a ChatMember.All filter which will allow all ChatMember updates. and it only makes sense for a ChatMember update handler to only handle chatMember updates.. or u can have it like this. In case the user's filter is nil. You should still filter out the ChatMember updates.

ext/handlers/filters/updates.go
JaskaranSM: There can be users who understand this structural "And Or" filter implementation better than the maze of anonymous functions. If you consider that situation, they are good.

ext/handlers/chatmember.go
PaulSonOfLars: I'm not sure i follow what you're saying, could you rephrase? 

In my mind, no filter (nil) == all chat_member updates. Because you haven't "filtered" out anything, so you would want ALL updates.

There arent any situations i can think of where "I dont want any updates" is ever a logical choice.

ext/handlers/filters/updates.go
PaulSonOfLars: But is that situation actually a real one?

Is it easier to read:
```go
func(u *gotgbot.ChatMemberUpdated) {
    return (filters.ChatMemberUserId(123)(u) && filters.ChatMemberChatId(567)(u)) || filters.ChatMemberChannel()(u)
}
```
or
```go
filters.ChatMemberUserId(123).And(filters.ChatMemberChatId(567)).Or(filters.ChatMemberChannel())
```

At least the operator precendence of the first one is clear. The second one is dubious (i think it's the same? not sure)

ext/handlers/chatmember.go
JaskaranSM: Dude, that ChatMemberAll filter func exists for a reason, does its name isnt able to make sense? And when did i ever say "i dont want any updates" part?

ext/handlers/filters/updates.go
JaskaranSM: The thing is i m fine with either of them, its your library, you should be the one deciding what to keep.

ext/handlers/chatmember.go
PaulSonOfLars: I'm trying to have a discussion with you. If you are unable to do that calmly and respectfully, please refrain from opening PRs in the future.

I dont see why the `ChatMemberAll` filter and a `nil` filter can't have the same functionality. As I have explained above, I dont think a `nil` filter should stop any handlers from executing, since none of the updates have been "filtered" out.

I wasn't quoting you. I was describing functionality of the current code, which I believe is the wrong choice, so was asking for **your opinions** on. Currently, if the filter is `nil`, that handler will never be triggered. AKA, "I dont want any updates". I don't see a situation where this might be a useful outcome, which is why I am arguing against it.

ext/handlers/filters/updates.go
PaulSonOfLars: I'm asking you for an opinion, and a use case. If you don't have an opinion, all you need to do is say so.

Just because its "my library" (which is not how i consider open source work), that doesnt mean I always have the correct opinions. Which is why I ask other people.

PaulSonOfLars (changes requested):
Id like to resolve the current discussions before merging this, since theyre important to the growth of the library.

ext/handlers/chatmember.go
JaskaranSM: > I dont see why the `ChatMemberAll` filter and a `nil` filter can't have the same functionality. As I have explained above, I dont think a `nil` filter should stop any handlers from executing, since none of the updates have been "filtered" out.

If you are already aware of it then why not just edit them in Suggested Changes?

PaulSonOfLars: Because I was asking for your opinion, and trying to understand your thought process on having it return as false.

This was a discussion, not a request. As I clearly stated in the opening comment.

ext/handlers/filters/updates.go
sylphiette: > But is that situation actually a real one?
> 
> Is it easier to read:
> 
> ```go
> func(u *gotgbot.ChatMemberUpdated) {
>     return (filters.ChatMemberUserId(123)(u) && filters.ChatMemberChatId(567)(u)) || filters.ChatMemberChannel()(u)
> }
> ```
> 
> or
> 
> ```go
> filters.ChatMemberUserId(123).And(filters.ChatMemberChatId(567)).Or(filters.ChatMemberChannel())
> ```
> 
> At least the operator precendence of the first one is clear. The second one is dubious (i think it's the same? not sure)

If I may, I think it's better to use golang operators, for both readability and ease of use. the And/Or/Not methods are for limited languages like java.

PaulSonOfLars: > If I may, I think it's better to use golang operators, for both readability and ease of use. the And/Or/Not methods are for limited languages like java.

Thanks for your input - I absolutely agree. I was looking for external opinions, so this is really appreciated. :)

I intend to refactor and remove them from all filters in a later PR, to keep things consistent

PR #23, opened 2021-05-24
Add ChosenInlineResult handler

rojvv:
# What

`ChosenInlineResult` update handler.

# Impact

- Have you included documentation, or updated existing documentation? -
- Do errors and log messages provide enough context? -

### DIFF
--- a/ext/handlers/filters/choseninlineresult.go
+++ b/ext/handlers/filters/choseninlineresult.go
@@ -1,0 +1,35 @@
+package filters
+
+import "github.com/PaulSonOfLars/gotgbot/v2"
+
+type ChosenInlineResult func(cir *gotgbot.ChosenInlineResult) bool
+
+func (f ChosenInlineResult) And(f2 ChosenInlineResult) ChosenInlineResult {
+	return func(cir *gotgbot.ChosenInlineResult) bool {
+		return f(cir) && f2(cir)
+	}
+}
+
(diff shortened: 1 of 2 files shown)


ext/handlers/filters/choseninlineresult.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+}
+
+func ChosenResultQuery(q string) ChosenInlineResult {
+	return func(cir *gotgbot.ChosenInlineResult) bool {
+		return cir.Query == q
+	}
+}
+
+func ChosenResultPrefix(prefix string) ChosenInlineResult {
```
PaulSonOfLars: Would it be better to call this `ChosenResultQueryPrefix` instead? Would make it easier to read the code and know what the prefix is for. What do you think?

PaulSonOfLars (commented):
Thank you for the contribution! 

Got one question regarding the naming of the prefix/suffix methods that id like your opinion on, otherwise looks great :)

ext/handlers/filters/choseninlineresult.go
PaulSonOfLars: (of course, i see this isnt the case for the current callback query filter; If you agree with the naming scheme, I'll rename that in a later commit)

rojvv: Done!

PR #27, opened 2021-06-28
Add Regex Filter for messages

celestix:
# What

Added regex filters to allow message handlers to match regex patterns.

### DIFF
--- a/ext/handlers/filters/message/message.go
+++ b/ext/handlers/filters/message/message.go
@@ -47,6 +47,16 @@ func ForwardFromChatID(id int64) filters.Message {
 	}
 }
 
+func Regex(pattern string) filters.Message {
+	return func(m *gotgbot.Message) bool {
+		matched, err := regexp.MatchString(pattern, m.Text)
+		if err != nil {
+			return false
+		}
+		return matched
+	}
+}
+
 func Reply(msg *gotgbot.Message) bool {
 	return msg.ReplyToMessage != nil
 }
--- a/ext/handlers/filters/message/message.go
+++ b/ext/handlers/filters/message/message.go
@@ -2,7 +2,6 @@ package message
 
 import (
 	"strings"
-	"log"
 	"regexp"
 
 	"github.com/PaulSonOfLars/gotgbot/v2"
@@ -49,16 +48,15 @@ func ForwardFromChatID(id int64) filters.Message {
 	}
 }
 
-func Regex(p string) filters.Message {
+func Regex(p string) (filters.Message, error) {
+	r, err := regexp.Compile(p)
+	if err != nil {
+		return nil, err
+	}
 	return func(m *gotgbot.Message) bool {
-		r, err := regexp.Compile(p)
-		if err != nil {
-			log.Fatal(err)
-			return false
-		}
 		matched := r.MatchString(m.Text)
 		return matched
-	}
+	}, nil
 }
 
 func Reply(msg *gotgbot.Message) bool {
--- a/ext/handlers/filters/message/message.go
+++ b/ext/handlers/filters/message/message.go
@@ -54,8 +54,7 @@ func Regex(p string) (filters.Message, error) {
 		return nil, err
 	}
 	return func(m *gotgbot.Message) bool {
-		matched := r.MatchString(m.Text)
-		return matched
+		return r.MatchString(m.Text)
 	}, nil
 }
 
--- a/ext/handlers/filters/message/message.go
+++ b/ext/handlers/filters/message/message.go
@@ -2,6 +2,8 @@ package message
 
 import (
 	"strings"
+	"log"
+	"regexp"
 
 	"github.com/PaulSonOfLars/gotgbot/v2"
 	"github.com/PaulSonOfLars/gotgbot/v2/ext/handlers/filters"
@@ -47,12 +49,14 @@ func ForwardFromChatID(id int64) filters.Message {
 	}
 }
 
-func Regex(pattern string) filters.Message {
+func Regex(p string) filters.Message {
 	return func(m *gotgbot.Message) bool {
-		matched, err := regexp.MatchString(pattern, m.Text)
+		r, err := regexp.Compile(p)
 		if err != nil {
+			log.Fatal(err)
 			return false
 		}
+		matched := r.MatchString(m.Text)
 		return matched
 	}
 }


ext/handlers/filters/message/message.go
```diff
@@ -47,6 +47,16 @@ func ForwardFromChatID(id int64) filters.Message {
 	}
 }
 
+func Regex(pattern string) filters.Message {
+	return func(m *gotgbot.Message) bool {
+		matched, err := regexp.MatchString(pattern, m.Text)
```
PaulSonOfLars: Regexes that are reused should be compiled for speed, see `regex.Compile`

ext/handlers/filters/message/message.go
```diff
@@ -47,6 +47,16 @@ func ForwardFromChatID(id int64) filters.Message {
...
+		if err != nil {
+			return false
```
PaulSonOfLars: This swallows all errors, making it very hard to debug. How do you know your match failed?

One option would be to change the definition to `Regex(p string) (filters.Message, error)`. This stops users from being able to instantiate the filter inline, but still allows them to handle the error.

PaulSonOfLars (changes requested):

ext/handlers/filters/message/message.go
ALiwoto: Agree

ext/handlers/filters/message/message.go
celestix: I see, going to try it out today. :) Thanks for the brief explanation.

ext/handlers/filters/message/message.go
celestix: Okay, Understood.

ext/handlers/filters/message/message.go
ALiwoto: Why not changing the type `filters.Message` from `func(*gotgbot.Message) bool` to `func(*gotgbot.Message) (bool, error)`?
Isn't it a good idea?

celestix:
Just committed both the changes. once have a look @PaulSonOfLars

ext/handlers/filters/message/message.go
```diff
@@ -47,12 +49,14 @@ func ForwardFromChatID(id int64) filters.Message {
 	}
 }
 
-func Regex(pattern string) filters.Message {
+func Regex(p string) filters.Message {
 	return func(m *gotgbot.Message) bool {
-		matched, err := regexp.MatchString(pattern, m.Text)
+		r, err := regexp.Compile(p)
 		if err != nil {
+			log.Fatal(err)
```
ALiwoto: Oh god
no please don't do that

ALiwoto: `log.Fatal(err)` 
this work is completely wrong
if we get an error, the application will completely terminated
it's not called error handling at all

ext/handlers/filters/message/message.go
ALiwoto: ~@anonyindian  please check my pull request [here](https://github.com/anonyindian/gotgbot/pull/1).~

PaulSonOfLars: The compile error only ever happens once (at compile time). The error can also only ever be handled at compile time, so there's no way to action any errors that would get returned with (bool, error) - so, what do you gain by returning the error?

ext/handlers/filters/message/message.go
```diff
@@ -47,6 +49,18 @@ func ForwardFromChatID(id int64) filters.Message {
 	}
 }
 
+func Regex(p string) filters.Message {
+	return func(m *gotgbot.Message) bool {
+		r, err := regexp.Compile(p)
+		if err != nil {
+			log.Fatal(err)
```
PaulSonOfLars: This will cause the program to panic. That is definitely not the intention here. The programmer needs to handle the error that gets returned - if they don't, that's their choice, but killing the program isn't the answer.

That's why I suggested that the `Regex` return both a message, and an error.

ext/handlers/filters/message/message.go
```diff
@@ -47,6 +49,18 @@ func ForwardFromChatID(id int64) filters.Message {
 	}
 }
 
+func Regex(p string) filters.Message {
+	return func(m *gotgbot.Message) bool {
+		r, err := regexp.Compile(p)
```
PaulSonOfLars: The compile should be moved outside of the anonymous `func()`, otherwise you recompile the regex every time and lose all the performance improvements from compiling it in the first place.

ext/handlers/filters/message/message.go
```diff
@@ -47,6 +49,18 @@ func ForwardFromChatID(id int64) filters.Message {
...
+		if err != nil {
+			log.Fatal(err)
+			return false
+		}
+		matched := r.MatchString(m.Text)
+		return matched
```
PaulSonOfLars:
```suggestion
	  return r.MatchString(m.Text)
```

No point having a temporary variable here

PaulSonOfLars (changes requested):

ALiwoto (commented):
> so, what do you gain by returning the error?
@PaulSonOfLars  What should be done then? Instead of returning the error?

ext/handlers/filters/message/message.go
ALiwoto: Oh right, this way if we got an error from `regexp.Compile`, we can return it by `Regex` function
so the `Regex` function will become something like this `func Regex(p string) (filters.Message, error)`
this way users can handle it directly

celestix: Right, Let me do so.

celestix:
@PaulSonOfLars Please review the last commit!

PaulSonOfLars (approved):
LGTM % last comment to remove an unnecessary temp variable

ALiwoto (approved):

celestix:
> LGTM % last comment to remove an unnecessary temp variable

Once check please, It should be fine now.

ext/handlers/filters/message/message.go
```diff
@@ -54,8 +54,7 @@ func Regex(p string) (filters.Message, error) {
 		return nil, err
 	}
 	return func(m *gotgbot.Message) bool {
-		matched := r.MatchString(m.Text)
-		return matched
+		return r.MatchString(m.Text)
```
ALiwoto, line 57: </img> nice

ALiwoto (approved):

PR #29, opened 2021-07-17
Custom mux (router) when starting webhook server

Camelva:
# What

Briefly and clearly: additional `Mux` field to `WebhookOpts`. With this we can register additional route handlers without need to start another `http.Server` instance.

# Impact


PaulSonOfLars:
Hey @Camelva, thank you for the great PR!

This is something I've definitely thought about in the past, and have been delaying until someone raised a request for it so that we can get a conversation going.

My thought process is as such:
If someone is passing their own mux, they're looking to re-use the server. However, it seems to me like it would be much more stable for them to own that server themselves, so any changes in the gotgbot lib (which would be fine for the lib, but breaking for other usages) dont affect them. In that case, it might actually be best for the gotgbot handler to  be run on THEIR server, with their own configs, instead of their mux running on the library server.

To get to the point; instead of allowing the user to pass a custom mux, what do you think of creating a helper method which creates the handler on an existing mux, leaving the server configuration to the developer instead of using the library defaults?

So to cover your use case (if i understand correctly), someone might have:
```golang
mux := http.NewServeMux()

// add all your own custom paths
mux.HandleFunc(...)

// New method to add the gotgbot handler mux to a path of your choice
u.HandleOnMux(mux, "/"+b.Token)

// Existing server, maintained by dev, not lib
s := http.Server{Handler: mux}

s.ListenAndServer(...)

```

What do you think of this approach?

(NB: Its very hot here, so this message might not be as clear as I would like it to be - dont hesitate to ask for clarification, or be critical)

Camelva:
Hey, @PaulSonOfLars, thanks for reply.

I thought about similar approach (but at that moment I thought about lib returning mux / server, instead of passing your own), and it raised additional questions, so i stopped with current, easier version.

Let's start with easier one. 
IMO, if I'm passing my own mux to function means i already registered all my handlers and want lib to take care about everything else. Simple and easy to use. And when i need to stop everything, i just call `updater.Stop()`. 

Now about second version (aka reverse approach) - "lib just registers handler and everything else is user's headache". 
It may be more versatile, but same time capable to create more __potential__ problems. Like, 
- what if user started server before starting `Dispatcher`?
- when user calls `updater.Stop()` server will keep running, right? This means user should manually stop server. And if next update occurs somewhere in-between closing `UpdateChan` and server shutting down it will cause panic. 
And maybe some more, but its all I could think about at this moment.

Conclusion: i think 2nd approach can be used as well, but as additional option, not the only one available. I also think it will be used very rarely. Like, **very**.

PaulSonOfLars:
**TL;DR**: If a user wants custom server logic, they should own their own server, not blindly trust that the library server "works".

>  want lib to take care about everything else

But why should it do "everything else"? The lib is there to provide a basic webhook setup if necessary - as soon as a custom mux with pre-existing handlers is added, then the server functionality is no longer "handles incoming webhooks", its suddenly "handles everything", which isn't the intention, or the goal. This results in the overall application being more fragile, as it relies on undefined behaviour (the server setup). Suddenly, making changes to the default server setup within the lib has to consider people overloading it for their own needs with a mux, which could cause unexpected breakages.

> but same time capable to create more potential problems.

Yes. That's the point. If a user wants to use advanced behaviour, then they should be aware of those cases, and able to handle them. I'm totally fine with people wanting to set up their own webhooks logic, and to "optimise" by only using one `http.Server`, but then that is now their responsibility to test and confirm that their code works; the library should not be responsible for any changes there.

> What if user started server before Dispatcher?

Then the user should read the docs before using advanced features, because they need to understand the application flow.

> This means user should manually stop server

Again, yes. If they choose to run their own server, it is their responsibility to stop the server before calling `updater.Stop()`. Thats the choice they make for using advanced features instead of "we handle this for you" features.


I think its important to keep track of what the responsibility of the lib is. I dont think running servers for external services, which could be set up in various strange and wonderful ways, is something that needs to be done by the lib.
Its also worth noting that the "configure this mux" approach that I've suggested could potentially make it possible to run multiple bots on the same `http.Server` in the future, since you would just register each bot handler onto the same mux without interacting with the running server.

What do you think?

Camelva:
Well, i feel like you overcomplicate this.

IMO passing your own mux when starting webhooks server needed to add some *utility* routes. For example, checking service health, exposing metrics or anything like that. IMO its still part of "bot" and still fits for lib's responsibility (e.g. health endpoint should only work if bot works).

For any "additional" features user can just start additional server instance and control it as they wish. If its not directly related to bot, its not lib's responsibility anymore (maybe some kind of web ui, idk. Can't think of anything else now lol).

For "advanced logic" of WebhookServer user can create their own `StartWebhook` function (and even their own `Updater`, if anything). All you need is `Dispatcher` (which is exported together with all important methods) and a channel to store updates.

As you said, helper method to create handler on an existing mux may be an option as well. But IMO its very close to "custom StartWebhook function" and fits for only 1 out of 3 cases i mentioned.

PR #30, opened 2021-08-01
Method to get message link

celestix:
Added Message.GetLink() Method to get message link

# What

Inspired by python-telegram-bot, added Message.GetLink() to make it easier to get the link of a message in a supergroup or channel.

### DIFF
--- a/custom_helpers.go
+++ b/custom_helpers.go
@@ -1,5 +1,22 @@
 package gotgbot
 
+import (
+	"fmt"
+	"strings"
+)
+
+/* GetLink is a helper method to easily get the message link excluding group and private chat types.
+(It will return en empty string in case of private or group chat type)*/
+func (m Message) GetLink() string {
+	if m.Chat.Type == "private" || m.Chat.Type == "group" {
+		return ""
+	}
+	if m.Chat.Username != "" {
+		return fmt.Sprintf("https://t.me/%s/%d", m.Chat.Username, m.MessageId)
+	}
+	return strings.Replace(fmt.Sprintf("https://t.me/c/%d/%d", m.Chat.Id, m.MessageId), "-100", "", 1)
+}
+
 // Reply is a helper function to easily call Bot.SendMessage as a reply to an existing message.
 func (m Message) Reply(b *Bot, text string, opts *SendMessageOpts) (*Message, error) {
 	if opts == nil {
--- a/custom_helpers.go
+++ b/custom_helpers.go
@@ -14,7 +14,9 @@ func (m Message) GetLink() string {
 	if m.Chat.Username != "" {
 		return fmt.Sprintf("https://t.me/%s/%d", m.Chat.Username, m.MessageId)
 	}
-	return strings.Replace(fmt.Sprintf("https://t.me/c/%d/%d", m.Chat.Id, m.MessageId), "-100", "", 1)
+	// Message links use raw chatIds without the -100 prefix; this trims that prefix.
+	rawChatId := strings.TrimPrefix(strconv.FormatInt(m.Chat.Id, 10), "-100")
+	return fmt.Sprintf("https://t.me/c/%s/%d", rawChatId, m.MessageId)
 }
 
 // Reply is a helper function to easily call Bot.SendMessage as a reply to an existing message.
--- a/custom_helpers.go
+++ b/custom_helpers.go
@@ -5,8 +5,7 @@ import (
 	"strings"
 )
 
-/* GetLink is a helper method to easily get the message link excluding group and private chat types.
-(It will return en empty string in case of private or group chat type)*/
+// GetLink is a helper method to easily get the message link (It will return an empty string in case of private or group chat type).
 func (m Message) GetLink() string {
 	if m.Chat.Type == "private" || m.Chat.Type == "group" {
 		return ""
--- a/gen_types.go
+++ b/gen_types.go
@@ -3562,6 +3562,20 @@ type Message struct {
 	ReplyMarkup *InlineKeyboardMarkup `json:"reply_markup,omitempty"`
 }
 
+// GetLink is a helper method to easily get the message link excluding group and private chat types.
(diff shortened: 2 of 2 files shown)


celestix:
I know that it will create an issue while creating link through chat id due to -100 prefix but I'm still here to clear a doubt that whether it will be ok to add this method or not.

gen_types.go
```diff
@@ -3562,6 +3562,20 @@ type Message struct {
 	ReplyMarkup *InlineKeyboardMarkup `json:"reply_markup,omitempty"`
 }
 
+// GetLink is a helper method to easily get the message link excluding group and private chat types.
```
PaulSonOfLars: Check the first line of this file. :) 

No methods should be manually added in `gen_*` files, as regenerating the API will remove manual changes. `custom_helpers` is probably the right place for this.

gen_types.go
```diff
@@ -3562,6 +3562,20 @@ type Message struct {
...
+func (m *Message) GetLink() string {
+	if m.Chat.Type != "private" && m.Chat.Type != "group" {
```
PaulSonOfLars: Consider reducing indentation levels by following the principle of "fail fast".
In this case, if the type is private, or group, return the empty string.

This reduces indentation, making your code easier to read.

gen_types.go
```diff
@@ -3562,6 +3562,20 @@ type Message struct {
...
+		var to_link string
+		if m.Chat.Username != "" {
+			to_link = m.Chat.Username
+		} else {
+			to_link = fmt.Sprintf("c/%v", m.Chat.Id)
+		}
+		return fmt.Sprintf("https://t.me/%s/%v", to_link, m.MessageId)
```
PaulSonOfLars: Nit: we know message ID is an int64, so using %d is better than %v

gen_types.go
```diff
@@ -3562,6 +3562,20 @@ type Message struct {
 	ReplyMarkup *InlineKeyboardMarkup `json:"reply_markup,omitempty"`
 }
 
+// GetLink is a helper method to easily get the message link excluding group and private chat types.
+func (m *Message) GetLink() string {
+	if m.Chat.Type != "private" && m.Chat.Type != "group" {
+		var to_link string
```
PaulSonOfLars: This variable adds an unnecessary extra layer of abstraction - it would be much simpler to use the previous "fail fast" approach and return the t.me string in two separate instances, one for public and the other for private supergroups, specifying two different `fmt.Sprintf` calls.

Also allows you to remove the extra indentation from the `else` statement. Win!

gen_types.go
PaulSonOfLars: Also worth mentioning that this method only returns links for channels and supergroups, and will return an empty string otherwise

PaulSonOfLars (commented):
Thank you for the PR! Definitely a good addition which makes sense to add here, since this functionality is pretty commonly used and hasn't got any drawbacks.

I've left some comments on general codestyle; I really need to hook up the linters I use locally into the CI for this repo so that I can automate these suggestions 😅 my bad for not having those yet, sorry.

gen_types.go
celestix: > Check the first line of this file. :)
> 
> No methods should be manually added in `gen_*` files, as regenerating the API will remove manual changes. `custom_helpers` is probably the right place for this.

Ah Sorry, My bad.

celestix: > Also worth mentioning that this method only returns links for channels and supergroups, and will return an empty string otherwise

Sure, I'll consider commenting it as well.

gen_types.go
celestix: Thanks for your valuable suggestions! I'll try to implement all the changes as much as I understood. :)

celestix (commented):
@PaulSonOfLars Please have a look on it now.

custom_helpers.go
```diff
@@ -1,5 +1,22 @@
...
+/* GetLink is a helper method to easily get the message link excluding group and private chat types.
+(It will return en empty string in case of private or group chat type)*/
+func (m Message) GetLink() string {
+	if m.Chat.Type == "private" || m.Chat.Type == "group" {
+		return ""
+	}
+	if m.Chat.Username != "" {
+		return fmt.Sprintf("https://t.me/%s/%d", m.Chat.Username, m.MessageId)
+	}
+	return strings.Replace(fmt.Sprintf("https://t.me/c/%d/%d", m.Chat.Id, m.MessageId), "-100", "", 1)
```
PaulSonOfLars:
```suggestion
	// Message links use raw chatIds without the -100 prefix; this trims that prefix.
	rawChatId := strings.TrimPrefix(strconv.FormatInt(m.Chat.Id, 10), "-100")
	return fmt.Sprintf("https://t.me/c/%s/%d", rawChatId, m.MessageId)
```

Small nit to ease readability. Too many statements on one line with undefined constants makes the code harder to read.
This updated version should help that and keep things shiny :) 

(you can hit the "commit suggestion" button above to add these changes right away)

custom_helpers.go
```diff
@@ -1,5 +1,22 @@
 package gotgbot
 
+import (
+	"fmt"
+	"strings"
+)
+
+/* GetLink is a helper method to easily get the message link excluding group and private chat types.
+(It will return en empty string in case of private or group chat type)*/
```
PaulSonOfLars: ultra nit: 
Golang doc generators tend to expect `//` style comments for function documentation. `/**/` comments are typically only used for in-code explanations of weird logic, or to comment out large blocks when debugging.

PaulSonOfLars (commented):
LGTM - two last small nits. 

Working on setting up linters/testrunners now, so we can skip my pedantics in the future 😅

custom_helpers.go
celestix: Oh, I was unaware about that, Sorry

celestix:
@PaulSonOfLars Please check now.

PaulSonOfLars (approved):

PR #37, opened 2021-10-28
Update types.go

Camelva:
# What
Small change to types generation script, so it will keep required fields in output insted of omitting them


PaulSonOfLars:
Hey @Camelva, thanks for the quick fix! 

Do you mind running `go generate` from the repo root to confirm this works and regenerate the type files?

Please commit and push them once done too so we can keep the lib in sync :)

Camelva:
done

PR #42, opened 2021-12-22
Paul/support dual returns

PaulSonOfLars:
# What

Support dual returns for methods such as `editMessageText`, which can return either the edited `Message`, or `bool`. 

Previously, this just returned Message, but this throws an error when the bool fails to unmarshal.
Instead, we now return both (and populate the boolean as true if the `msg != nil`). 

Example error: https://t.me/GotgbotChat/6877

# Impact

- Are your changes backwards compatible? No - this changes some function definitions. These are all compile-time errors, and simple to fix.
- Have you included documentation, or updated existing documentation? Yes.


ALiwoto (approved):
These changes look good to me, thanks for fixing the issue

PR #44, opened 2022-03-04
Add poll and pollanswer handler

celestix:
Implement poll and pollanswer handler for incoming respective updates.


PaulSonOfLars:
LGTM, thanks for the PR!

PR #55, opened 2022-06-23
Conversation handlers

PaulSonOfLars:
# What

Implement a new type of `Handler`, the conversation handler, which allows for storing the state of each user and emulate a "conversation" with them.

This does not add support for conversation timeouts or forced linear processing of conversation events; those can be added in the future, when needed.

# Impact

- Are your changes backwards compatible? No; change to the handler interface signature.


samples/conversationBot/main.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+// These are the conversation handler states.
+const (
+	NAME = "name"
+	AGE  = "age"
+)
+
+func main() {
+	// Get token from the environment variable
```
rojvv:
```suggestion
	// Get token from the environment variable.
```

dlgg-new:
The feature is very useful!  I appreciate that you developed such a project. I'd like to contribute to this project in the future.

PR #65, opened 2022-12-19
Replace error loggers in updater+dispatcher with error funcs, for improved flexibility

PaulSonOfLars:
# What

Improve error handling flexibility in the Updater and Dispatcher by providing error funcs instead of simple logs.

This should allow users to panic and close their program if necessary. 

# Impact

- Are your changes backwards compatible? No - some method and struct signatures changed to apply new changes. Overall minor. 

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -29,16 +29,13 @@ output:
 linters:
   disable-all: true
   enable:
-    - deadcode
     - errcheck
     - gosimple
     - govet
     - ineffassign
     - staticcheck
-    - structcheck
     - typecheck
     - unused
-    - varcheck
     - asciicheck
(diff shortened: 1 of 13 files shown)


ALiwoto:
I'm seriously waiting for these changes to be merged :)
Really good job!

.golangci.yaml
```diff
@@ -29,16 +29,13 @@ output:
 linters:
   disable-all: true
   enable:
-    - deadcode
```
ALiwoto, line 32: Can I ask why these should get disabled? (just curiosity)

PaulSonOfLars: They've been deprecated :) 
image

PR #69, opened 2022-12-31
Fix Updater.Stop() panic

vassilit:


PaulSonOfLars:
Good catch, thank you!

PR #73, opened 2023-01-07
add an option to listen on UNIX socket

vassilit:
For those that are running their bot behind a Web server, this MR contains a patch to listen on UNIX socket (as well as some small other unrelated changes that I can exclude from this MR if you don't want to merge them).


vassilit:
(backward-compatible, as this adds an optional field)

vassilit:
split out at the request of  the package maintainer.

PR #74, opened 2023-01-08
rework the web hook listen options

vassilit:
For those that are running their bot behind a Web server, this MR contains a patch to listen on UNIX socket.
(backward-compatible, as this adds an optional field)

### DIFF
--- a/ext/updater.go
+++ b/ext/updater.go
@@ -362,23 +362,18 @@ func (u *Updater) StartServer(opts WebhookOpts) error {
 		return ErrMissingCertOrKeyFile
 	}
 
-	var network string
-	if opts.UnixListen != "" {
-		network = "unix"
-	} else {
-		network = "tcp"
+	opts.setDefaults()
+
+	ln, err := net.Listen(opts.ListenNet, opts.ListenAddr)
+	if err != nil {
+		return fmt.Errorf("failed to listen on %s:%s: %w", opts.ListenNet, opts.ListenAddr, err)
 	}
 
 	u.webhookServer = &http.Server{
-		Addr:              opts.GetListenAddr(),
+		Addr:              ln.Addr().String(),
 		Handler:           u.serveMux,
 		ReadTimeout:       opts.ReadTimeout,
 		ReadHeaderTimeout: opts.ReadHeaderTimeout,
-	}
-
-	ln, err := net.Listen(network, u.webhookServer.Addr)
-	if err != nil {
-		return fmt.Errorf("failed to listen on %s:%s: %w", network, u.webhookServer.Addr, err)
 	}
 
 	go func() {
--- a/ext/updater.go
+++ b/ext/updater.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"net"
 	"net/http"
 	"strconv"
 	"strings"
@@ -352,12 +353,20 @@ func (u *Updater) StartServer(opts WebhookOpts) error {
 	}
 
 	var tls bool
-	if opts.CertFile == "" && opts.KeyFile == "" {
+	switch {
+	case opts.CertFile == "" && opts.KeyFile == "":
 		tls = false
-	} else if opts.CertFile != "" && opts.KeyFile != "" {
+	case opts.CertFile != "" && opts.KeyFile != "":
 		tls = true
-	} else {
+	default:
 		return ErrMissingCertOrKeyFile
+	}
+
+	var network string
+	if opts.UnixListen != "" {
+		network = "unix"
+	} else {
+		network = "tcp"
 	}
 
 	u.webhookServer = &http.Server{
@@ -367,12 +376,16 @@ func (u *Updater) StartServer(opts WebhookOpts) error {
 		ReadHeaderTimeout: opts.ReadHeaderTimeout,
 	}
 
+	ln, err := net.Listen(network, u.webhookServer.Addr)
+	if err != nil {
+		return fmt.Errorf("failed to listen on %s:%s: %w", network, u.webhookServer.Addr, err)
+	}
+
 	go func() {
-		var err error
 		if tls {
-			err = u.webhookServer.ListenAndServeTLS(opts.CertFile, opts.KeyFile)
+			err = u.webhookServer.ServeTLS(ln, opts.CertFile, opts.KeyFile)
 		} else {
-			err = u.webhookServer.ListenAndServe()
+			err = u.webhookServer.Serve(ln)
 		}
 		if err != nil && !errors.Is(err, http.ErrServerClosed) {
 			panic("http server failed: " + err.Error())
--- a/ext/updater.go
+++ b/ext/updater.go
@@ -362,15 +362,12 @@ func (u *Updater) StartServer(opts WebhookOpts) error {
 		return ErrMissingCertOrKeyFile
 	}
 
-	opts.setDefaults()
-
-	ln, err := net.Listen(opts.ListenNet, opts.ListenAddr)
+	ln, err := net.Listen(opts.GetListenNet(), opts.ListenAddr)
 	if err != nil {
 		return fmt.Errorf("failed to listen on %s:%s: %w", opts.ListenNet, opts.ListenAddr, err)
 	}
 
 	u.webhookServer = &http.Server{
-		Addr:              ln.Addr().String(),
 		Handler:           u.serveMux,
 		ReadTimeout:       opts.ReadTimeout,
 		ReadHeaderTimeout: opts.ReadHeaderTimeout,
--- a/ext/webhook.go
+++ b/ext/webhook.go
@@ -11,6 +11,8 @@ type WebhookOpts struct {
 	Listen string
 	// Port is the port listen on (eg 443, 8443, etc).
 	Port int
+	// UnixListen is the UNIX socket to listen on (mutually exclusive with Listen/Port options)
+	UnixListen string
 	// ReadTimeout is passed to the http server to limit the time it takes to read an incoming request.
 	// See http.Server for more details.
 	ReadTimeout time.Duration
@@ -29,6 +31,9 @@ type WebhookOpts struct {
 
 // GetListenAddr returns the local listening address, including port.
 func (w *WebhookOpts) GetListenAddr() string {
+	if w.UnixListen != "" {
+		return w.UnixListen
+	}
 	if w.Listen == "" {
 		w.Listen = "0.0.0.0"
 	}
--- a/ext/webhook.go
+++ b/ext/webhook.go
@@ -28,8 +28,9 @@ type WebhookOpts struct {
 	SecretToken string
 }
 
-func (w *WebhookOpts) setDefaults() {
+func (w *WebhookOpts) GetListenNet() string {
 	if w.ListenNet == "" {
-		w.ListenNet = "tcp"
+		return "tcp"
 	}
+	return w.ListenNet
 }
--- a/ext/webhook.go
+++ b/ext/webhook.go
@@ -1,18 +1,17 @@
 package ext
 
 import (
-	"fmt"
 	"time"
 )
 
 // WebhookOpts represent various fields that are needed for configuring the local webhook server.
 type WebhookOpts struct {
-	// Listen is the address to listen on (eg: localhost, 0.0.0.0, etc).
-	Listen string
-	// Port is the port listen on (eg 443, 8443, etc).
-	Port int
-	// UnixListen is the UNIX socket to listen on (mutually exclusive with Listen/Port options)
-	UnixListen string
+	// ListenAddr is the address and port to listen on (eg: localhost:http, 0.0.0.0:8080, :https, "[::1]:", etc).
+	// See the net package for details.
+	ListenAddr string
+	// ListenNet is the network type to listen on (must be "tcp", "tcp4", "tcp6", "unix" or "unixpacket").
+	// Empty means the default, "tcp".
+	ListenNet string
 	// ReadTimeout is passed to the http server to limit the time it takes to read an incoming request.
 	// See http.Server for more details.
 	ReadTimeout time.Duration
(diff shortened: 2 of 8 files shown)


ext/updater.go
```diff
@@ -352,27 +353,41 @@ func (u *Updater) StartServer(opts WebhookOpts) error {
...
 		Handler:           u.serveMux,
 		ReadTimeout:       opts.ReadTimeout,
 		ReadHeaderTimeout: opts.ReadHeaderTimeout,
 	}
 
+	var err error
+	var ln net.Listener
+	if ln, err = net.Listen(network, u.webhookServer.Addr); err != nil {
+		panic("listening on " + network + ":" + u.webhookServer.Addr + " failed: " + err.Error())
+	}
```
PaulSonOfLars: This inlined if statement doesn't really sense given all the values are used outside of it, and it should return an error rather than a panic.
Also, we should make sure to close the network with a defer.

```suggestion
	ln, err := net.Listen(network, u.webhookServer.Addr);
	if  err != nil {
		return fmt.Errorf("failed to listen on %s:%s: %w", network, u.webhookServer.Addr, err)
	}
	
	defer ln.Close()
```

ext/webhook.go
```diff
@@ -11,6 +11,8 @@ type WebhookOpts struct {
 	Listen string
 	// Port is the port listen on (eg 443, 8443, etc).
 	Port int
+	// UnixListen is the UNIX socket to listen on (mutually exclusive with Listen/Port options)
```
PaulSonOfLars: Could we throw some errors in the case where both unixListen AND Listen/Port are specified?

PaulSonOfLars: Or, alternatively - we could even add a "NetworkType" field to `WebhookOpts` which:

if empty or "tcp", uses TCP
else, uses the specified network (so unix would be supported through that).

We can then reuse the `Listen` field, and avoid having to duplicate `Listen` + `UnixListen`, including handling undefined behaviour of having both included.

What do you think? I think this would make it much easier to extend and support other network types in the future, and allows us to trickle up that logic to the caller rather than the implementation :)

ext/updater.go
```diff
@@ -352,27 +353,41 @@ func (u *Updater) StartServer(opts WebhookOpts) error {
...
 		return ErrMissingCertOrKeyFile
 	}
 
+	var network string
+	if opts.UnixListen != "" {
+		network = "unix"
+	} else {
+		network = "tcp"
+	}
+
```
PaulSonOfLars: See my suggestion on using a `NetworkType` below, so we don't need to infer types based on other fields.

Otherwise, I would prefer this kind of syntax, which IMO more clearly shows the default as TCP :) 
```
	network := "tcp"
	if opts.UnixListen != "" {
		network = "unix"
	}

```

PaulSonOfLars (changes requested):
Hey, thank you for your contribution! This is really valuable stuff, and I can definitely see the use case for it. Just have a small few comments on how best to expose this to the user, and then we're good to go. Happy to do a release once this is in too.

I'm not personally familiar with how to use unix sockets for this kind of stuff. If possible, would you mind writing a quick description (Either in a sample bot, or just in your PR for future reference) of how to use sockets locally, or on webserver?

ext/updater.go
vassilit: > This inlined if statement doesn't really sense

I agree

> Also, we should make sure to close the network with a defer.

`defer ln.Close()` is wrong here, as in the next step we are spawning a goroutine that uses `ln` and then immediately `return nil`, thus executing the deferred `ln.Close()`. We need `ln` to stay open for `Serve()` to work.

Also, we should not add a `defer ln.Close()` in the anonymous function, because `http.Serve()` does it and already go to great lengths to avoid closing the listener twice (see [code](https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/net/http/server.go;l=3488;drc=f721fa3be9bb52524f97b409606f9423437535e8;bpv=0;bpt=0) and issue golang/go#24803).

ext/webhook.go
vassilit: I don't have in mind any other «network type» (transport layer) for HTTP that could be useful here, apart maybe "sctp".

Alternatively we could parse the `Listen` field for the `/` character, as it is invalid in IPv4 and IPv6 addresses notation as well as in DNS hostnames. That would however force to write `./tgbot.sock` instead of just `tgbot.sock` if referring to a socket in the current working directory.

vassilit:
Example for testing with ngrok:
```
ngrok http 8080
socat TCP-LISTEN:8080 UNIX-CONNECT:/tmp/smtgbot.sock
```

ext/updater.go
vassilit: > it should return an error rather than a panic.

Agree. I was influenced by the later `panic("http server failed: " + err.Error())`, but here the situation is different, we still have the chance to return an error.

ext/updater.go
vassilit: I think you're right, this is too convoluted. I was trying to be backward-compatible, but since we are still in rc, I think we should:

- remove the `Port` and `UnixListen` fields and add a `ListenNet` field.
- `ListenNet` can take "tcp", "tcp4", tcp6", "unix", "unixpacket" values. If empty, assume "tcp".
- the `Listen` field uses a "host:port" or a path syntax, depending on `ListenNet`.

ext/updater.go
```diff
@@ -352,27 +353,34 @@ func (u *Updater) StartServer(opts WebhookOpts) error {
...
+	opts.setDefaults()
+
+	ln, err := net.Listen(opts.ListenNet, opts.ListenAddr)
+	if err != nil {
+		return fmt.Errorf("failed to listen on %s:%s: %w", opts.ListenNet, opts.ListenAddr, err)
+	}
+
 	u.webhookServer = &http.Server{
-		Addr:              opts.GetListenAddr(),
+		Addr:              ln.Addr().String(),
```
vassilit: Actually, as webhookServer is not exported we could skip this line completely.
srv.Addr is not used at all by the http.Server (except for srv.ListenAndServe).

ext/updater.go
```diff
@@ -352,27 +353,34 @@ func (u *Updater) StartServer(opts WebhookOpts) error {
...
+	case opts.CertFile != "" && opts.KeyFile != "":
 		tls = true
-	} else {
+	default:
 		return ErrMissingCertOrKeyFile
 	}
 
+	opts.setDefaults()
+
+	ln, err := net.Listen(opts.ListenNet, opts.ListenAddr)
```
vassilit: if ListenNet or/and ListenAddr are not properly formatted, net.Listen will return a complex error message indicating what part it doesn't like (those two strings get a lot of parsing in the std lib anyway, so we should delegate this part to them).
But if you want, we could add `errors.Is()` checks on wrapped `net.UnknownNetworkError`, `net.InvalidAddrError`, `net.DNSError`  and `net.AddrError` and show a message like "something is wrong with your address/network, check your configuration".

samples/echoWebhookBot/go.mod
```diff
@@ -2,6 +2,6 @@ module github.com/PaulSonOfLars/gotgbot/samples/echoWebhookBot
 
 go 1.15
 
-require github.com/PaulSonOfLars/gotgbot/v2 v2.0.0-rc.8
+require github.com/PaulSonOfLars/gotgbot/v2 v2.99.99
```
vassilit, line 5: the `replace` directive below takes whatever version is checked out in the parent directory, so I'd put a fake version here to avoid confusion.

PaulSonOfLars: A fair point, and removes the need for fixing it every so often! thanks.

samples/echoWebhookBot/main.go
```diff
@@ -70,8 +70,7 @@ func main() {
 	// Start the webhook server. We start the server before we set the webhook itself, so that when telegram starts
 	// sending updates, the server is already ready.
 	webhookOpts := ext.WebhookOpts{
-		Listen:      "0.0.0.0", // This example assumes you're in a dev environment running ngrok on 8080.
-		Port:        8080,
+		ListenAddr:  ":http-alt",   // This example assumes you're in a dev environment running ngrok on 8080.
```
PaulSonOfLars: Do you know how common the `:http-alt` syntax is? I've personally never seen it before, and would probably be more comfortable using `:8080`, if thats OK with you. I've also re-added the "0.0.0.0", to be explicit about what the example listening on; makes it clearer to anyone reading it imo, what do you think? :) 

```suggestion
		ListenAddr:  "0.0.0.0:8080",   // This example assumes you're in a dev environment running ngrok on 8080.
```

ext/updater.go
PaulSonOfLars: I think we should be fine returning like it is for now; we can change this in the future if people seem to be getting confused.

ext/updater.go
PaulSonOfLars: Good spot, I'd be up for removing it if you are! No point having dead code in there.

ext/webhook.go
```diff
@@ -27,13 +28,8 @@ type WebhookOpts struct {
 	SecretToken string
 }
 
-// GetListenAddr returns the local listening address, including port.
-func (w *WebhookOpts) GetListenAddr() string {
-	if w.Listen == "" {
-		w.Listen = "0.0.0.0"
+func (w *WebhookOpts) setDefaults() {
```
PaulSonOfLars: What's more idiomatic; a `setDefaults` method, or a `GetListenNet` which handles the default case? 

I think I lean towards the second; I've been bitten by methods unexpectedly changing their input pointers before. What do you think?

PaulSonOfLars (commented):
Couple more small comments. Sorry for the many review iterations; this is my first time doing a code review in a while, and I find these changes really exciting :)

samples/echoWebhookBot/go.mod
vassilit: Even if it doesn't change anything, setting this to the minimum acceptable version compatible with the program could be a good practice. If the sample program don't use a changed API, we don't update this.

Also, to work with multiple `go.mod` in a single repository: https://go.dev/doc/tutorial/workspaces

samples/echoWebhookBot/main.go
vassilit: `0.0.0.0:8080` is good. But we should document that if the host in the address parameter is empty, we listen on all available unicast and anycast IP addresses of the local system (this is what `net.Listen()` does).
Or even better, I think we would default to `localhost:8080` if we assume that we are in a dev env, no need to listen on all interfaces.

ext/webhook.go
vassilit: Not sure I understand about changing input pointers. `Webhookopts` stays at the same address, only the string changes.

About `GetListenNet`, could write a suggestion code ? I'm not sure to understand how you would use it.

samples/echoWebhookBot/main.go
PaulSonOfLars: Sure, that seems fair to me. Localhost should be enough, you're right!

samples/echoWebhookBot/go.mod
PaulSonOfLars: Yeah, fair.

Re: workspaces; I'm aware of them, I had simply opted to use the require here because i wanted to point out that this isnt intended to be a working go.mod file. If I understand it right, even if we used a workspace, we would still have to use the dummy versions you mentioned here, so wouldn't really solve much :)

ext/webhook.go
PaulSonOfLars: What I was trying to say is that by calling `setDefaults`, we end up changing the `opts` parameter in `StartServer`.

So, whoever calls `StartServer` has to be aware of the side effect that it will mutate the opts parameter it passed in, which seems like an unnecessary side effect.

Re: code, I just meant something like this, so the empty case is handled instead of defaulted.

```
func (w *WebhookOpts) GetListenNet() string {
	if w.ListenNet == "" {
		return "tcp"
	}
	return w.ListenNet
```

And then you would instead call `net.Listen(opts.GetListenNet()`, ...)` when starting the server.

vassilit: > So, whoever calls StartServer has to be aware of the side effect that it will mutate the opts parameter it passed in, which seems like an unnecessary side effect.

It won't, `opts` is passed by value to `StartServer`, so the the the caller's reference to its `opts.ListenNet` will stay the same, an empty string.

Every non-zero member (that's two 64-bit word per string, a pointer and a length) of it is pushed to the stack (would be in registers if the structure was smaller) twice: once for reuse from main, and once for calling StartServer. Once StartServer returns, main uses the original copy:

```
echoMultiBot:
(main.go:122) LEAQ    go.string."127.0.0.1:8080"(SB), DX  ; here it stores ListenAddr's addr into DX
(main.go:122) MOVQ    DX, main.opts+200(SP)    ; store this address in the stack 
(main.go:122) MOVQ    $14, main.opts+208(SP)  ; store 14 (the length of ListenAddr) in the stack
(main.go:123) MOVQ    R9, main.opts+280(SP)    ; store SecretToken's addr in the stack
(main.go:123) MOVQ    R10, main.opts+288(SP)  ; idem, the len of SecretToken
[...]
(main.go:127) MOVQ    DX, (SP)      ; move a SECOND copy of ListenAddr's addr into the stack
(main.go:127) MOVQ    $14, 8(SP)  ; the length stay the same, move it again to another addr in the stack
(main.go:127) MOVQ    R9, 80(SP)
(main.go:127) MOVQ    R10, 88(SP)
(main.go:127) PCDATA  $1, $1
(main.go:127) CALL    gotgbot/v2/ext.(*Updater).StartServer(SB)
[...]
; reuse our original unmodified opts to call AddWebhook
```

PaulSonOfLars: Argh, too used to all the other opts calls. Thought that one was passed as a pointer too, and didn't check to confirm.

Regardless, I still think it's nicer to avoid this pattern of "setting defaults", which could bite you if you forget to call it.

vassilit: In C, with an explicit `const` qualifier, it's easier to take a pointer while telling the user that we won't modify anything (even if it's not enforced per se, one can always cast const away).

PaulSonOfLars (approved):
Awesome, LGTM! Thanks for the contribution :)

PR #75, opened 2023-01-08
simplify signal-only channels

vassilit:
We are not using the boolean property in our semaphore, so we should use `chan struct{}` instead of `chan bool`.


PaulSonOfLars (approved):
Looks good, thanks! Was not aware that struct{} could be used to avoid allocations on channels, very cool.

Also a good use of channel closes, learnt something there too!

ALiwoto:
I see

annihilatorrrr:
Cool!

PR #76, opened 2023-01-08
avoid unnecessary Sprintf

vassilit:
`Sprintf` should be less efficient than `Join`, as we are effectively joining two strings by a separator.
The «proper» way could have been to use the `net/url` package, but `Join` is simpler.

### DIFF
--- a/ext/updater.go
+++ b/ext/updater.go
@@ -331,7 +331,7 @@ func (u *Updater) AddWebhook(b *gotgbot.Bot, urlPath string, opts WebhookOpts) {
 // SetAllBotWebhooks sets all the webhooks for the bots that have been added to this updater via AddWebhook.
 func (u *Updater) SetAllBotWebhooks(domain string, opts *gotgbot.SetWebhookOpts) error {
 	for _, data := range u.botMapping {
-		_, err := data.bot.SetWebhook(fmt.Sprintf("%s/%s", strings.TrimSuffix(domain, "/"), data.urlPath), opts)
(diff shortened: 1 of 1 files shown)


PaulSonOfLars (changes requested):
You're right, let's just use `url.JoinPath` here; that seems much cleaner, and removes the need for the `strings.Suffix`!

I guess it could cause issues if the `data.urlPath` contained extra slashes; but I think we can cross that bridge when we get to it, seems safe enough for now. What do you think?

vassilit:
> You're right, let's just use url.JoinPath here; that seems much cleaner, and removes the need for the strings.Suffix!

Yes, but we should modify go.mod also, because `url.JoinPath` appeared in Go 1.19.

> I guess it could cause issues if the data.urlPath contained extra slashes

if `data.urlPath` contain extra slashes right now, this is already a problem, as we don't know how the telegram server will handle ".././../" in URLs. We could use `url.PathEscape` on the urlPath part.

PaulSonOfLars:
> Yes, but we should modify go.mod also, because `url.JoinPath` appeared in Go 1.19.

Ahh, thats why. Makes sense, guess we should wait on that for a bit then. Shame.

> if `data.urlPath` contain extra slashes right now, this is already a problem, as we don't know how the telegram server will handle ".././../" in URLs. We could use `url.PathEscape` on the urlPath part.

True. Eh, I'm OK with calling that user error. We probably shouldnt mess with that input too much, since itll be used by the user in other places.

PaulSonOfLars (approved):
LGTM, but good to note we can update to url.JoinPath when the module gets upgraded to 1.19

PR #77, opened 2023-01-08
optimisation of GC for heavily-accessed structures and small linting

vassilit:
The Go garbage collector scans for pointers to free referenced objects in every allocated structures up to the last field that potentially reference something.

By moving reference field to the beginning of the structure this allows the GC to fetch less often from memory as they can fetch the whole referenced part of the structure in one operation if it fits the CPU cacheline size

For example, `handlers.Conversation` became a consolidated 48 bytes of pointers instead of 72 bytes. If we follow 10 000 conversations this may well reduce the GC interruption time.

Of course, they are numerous factors (like a change of the Go runtime GC implementation in the future) that can make this optimisation useless, but as far as Go 1.19, there is no harm to try.

Others seldom allocated structures are not pointer-aligned, but as changing their order may break users using unnamed fields, better not to change them.

### DIFF
--- a/ext/dispatcher.go
+++ b/ext/dispatcher.go
@@ -196,7 +196,6 @@ func (d *Dispatcher) Start(b *gotgbot.Bot, updates chan json.RawMessage) {
 					d.logf("Failed to process update: %s", err.Error())
 				}
 			}
-
 		}(upd)
 	}
 }
@@ -243,13 +242,13 @@ func (d *Dispatcher) ProcessUpdate(b *gotgbot.Bot, update *gotgbot.Update, data 
 			// If a panic handler is defined, handle the error.
 			if d.Panic != nil {
 				d.Panic(b, ctx, r)
-				return
 
-			} else {
-				// Otherwise, create an error from the panic, and return it.
-				err = fmt.Errorf("%w: %v\n%s", ErrPanicRecovered, r, cleanedStack())
 				return
 			}
+			// Otherwise, create an error from the panic, and return it.
+			err = fmt.Errorf("%w: %v\n%s", ErrPanicRecovered, r, cleanedStack())
+
+			return
 		}
 	}()
 
--- a/ext/handlers/conversation.go
+++ b/ext/handlers/conversation.go
@@ -36,14 +36,14 @@ type Conversation struct {
 }
 
 type ConversationOpts struct {
+	// StateStorage is responsible for storing all running conversations.
+	StateStorage conversation.Storage
 	// Exits is the list of handlers to exit the current conversation partway (eg /cancel commands)
 	Exits []ext.Handler
 	// Fallbacks is the list of handlers to handle updates which haven't been matched by any states.
 	Fallbacks []ext.Handler
 	// If True, a user can restart the conversation by hitting one of the entry points.
 	AllowReEntry bool
-	// StateStorage is responsible for storing all running conversations.
-	StateStorage conversation.Storage
 }
 
 func NewConversation(entryPoints []ext.Handler, states map[string][]ext.Handler, opts *ConversationOpts) Conversation {
@@ -122,10 +122,10 @@ func (c Conversation) HandleUpdate(b *gotgbot.Bot, ctx *ext.Context) error {
 type ConversationStateChange struct {
 	// The next state to handle in the current conversation.
 	NextState *string
-	// End the current conversation
-	End bool
 	// Move the parent conversation (if any) to the desired state.
 	ParentState *ConversationStateChange
+	// End the current conversation
+	End bool
 }
 
 func (s *ConversationStateChange) Error() string {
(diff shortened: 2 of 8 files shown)


ext/dispatcher.go
```diff
@@ -243,13 +242,13 @@ func (d *Dispatcher) ProcessUpdate(b *gotgbot.Bot, update *gotgbot.Update, data
...
 				d.Panic(b, ctx, r)
-				return
 
-			} else {
-				// Otherwise, create an error from the panic, and return it.
-				err = fmt.Errorf("%w: %v\n%s", ErrPanicRecovered, r, cleanedStack())
 				return
 			}
+			// Otherwise, create an error from the panic, and return it.
+			err = fmt.Errorf("%w: %v\n%s", ErrPanicRecovered, r, cleanedStack())
```
PaulSonOfLars, line 249: This looks like a logic change. Previously, the Panic didn't set the error (as the idea was that it had been handled).

Is this intentional? If yes, what is the reasoning behind it?

ext/handlers/common_test.go
```diff
@@ -33,11 +33,11 @@ func NewTestBot() *gotgbot.Bot {
 	}
 }
 
-func NewMessage(userId int64, chatId int64, message string) *ext.Context {
+func NewMessage(userId, chatId int64, message string) *ext.Context {
```
PaulSonOfLars: TL;DR Specifying the type after each parameter is intentional, and I wouldn't want to remove them.

IMO, skipping the types might reduce that character count, but it makes the method harder to read - you have to look further than you variable to check.
It also opens up to new bugs, such as if someone were to add a parameter, and unknowingly change other parameter types.

eg `NewMessage(userId, newfield string, chatId int64, message string)` 

Suddenly, userId is a string instead of an int64. 

This would get caught on internal methods with compile errors, but might not be caught on external methods; causing issues down the line.

PaulSonOfLars (changes requested):
I have to admit that I think the performance benefits of the struct moves are minimal here. You would really need to allocate a LOT of callback/conversation handlers before struct alignment ever became an issue. I'm sure there would be other things to fix before that.

Regardless, I appreciate that you looked into this, and I'm happy to merge in the struct changes once the other comments are addressed :)

ext/handlers/common_test.go
vassilit: Agree, I will remove those changes (although they would probably get caught on external methods as well because we use that variable).

ext/dispatcher.go
vassilit: I've just changed 
```
if something {
    // some code
    return
} else {
    // some code
    return
}
```
by
```
if something {
    // some code
    return
}
// some code
return
```

AFAIK, no logic change.

PaulSonOfLars: Ah - yes, my bad! Misread the git diff and didnt see that the second return wasnt removed. You're entirely right!

ext/handlers/conversation.go
```diff
@@ -36,14 +36,14 @@ type Conversation struct {
...
+	// StateStorage is responsible for storing all running conversations.
+	StateStorage conversation.Storage
 	// Exits is the list of handlers to exit the current conversation partway (eg /cancel commands)
 	Exits []ext.Handler
 	// Fallbacks is the list of handlers to handle updates which haven't been matched by any states.
 	Fallbacks []ext.Handler
 	// If True, a user can restart the conversation by hitting one of the entry points.
 	AllowReEntry bool
-	// StateStorage is responsible for storing all running conversations.
-	StateStorage conversation.Storage
```
PaulSonOfLars, line 46: So, you mention that:
> For example, handlers.Conversation became a consolidated 48 bytes of pointers instead of 72 bytes. If we follow 10 000 conversations this may well reduce the GC interruption time.

By "follow 10 000 conversations", do you mean having a conversation with 10 000 users? Is GC interruption time an issue for anyone at the moment...?

Because this struct is only used when the conversation itself is created; eg, in this case, it is used exactly as many times as `dispatcher.AddHandler(handlers.NewConversation(...))` is called. Which, for what I expect to be the VAST majority of cases, would only be once, at startup. Given the nature of this libary, things are going to be primarily network IO bound, so optimising a one-time call seems a little overkill.

I'm happy to merge the PR, but I'm just wary of these calculations which seem to be suspiciously close to premature optimisation based on a misunderstanding of the struct function 😅

vassilit: Indeed, I haven't looked enough at the code. Just the name "Conversation" felt like it was referring to a conversation, not a conversation handler. A conversation now is essentially a single-element structure, State, that doesn't need optimisation at the moment. 

Anyway, it was not structure alignement (to save space in memory) but references in structure grouping (to save CPU time during every GC). But if there are only a few instances of this structure, we can abandon this patch indeed.

vassilit:
«Premature optimization is the root of all evil»

ext/handlers/conversation.go
PaulSonOfLars: No worries at all! 

This is not an aspect I often focus on when writing code, so I very much appreciate your input here, and for taking the time to answer my questions/doubts. Thank you for looking into this!

PR #79, opened 2023-01-09
Add support for auto-generating ChatTypes constants.

ALiwoto:
# What
Since we already have [UpdateType](https://github.com/PaulSonOfLars/gotgbot/blob/371f60d4841ef41c6b99542f3157f97be45ca5c9/gen_consts.go#L7), [ParseMode](https://github.com/PaulSonOfLars/gotgbot/blob/371f60d4841ef41c6b99542f3157f97be45ca5c9/gen_consts.go#L25) and [StickerType](https://github.com/PaulSonOfLars/gotgbot/blob/371f60d4841ef41c6b99542f3157f97be45ca5c9/gen_consts.go#L33) constants, I thought it's better we have `ChatType` constants as well (this makes it easy to do comparison on `Chat.Type` field on custom filters/handlers for us)

# Impact

- Are your changes backwards compatible? Yes.
- Have you included documentation, or updated existing documentation? Yes.
- Do errors and log messages provide enough context? Yes.

### DIFF
--- a/scripts/generate/consts.go
+++ b/scripts/generate/consts.go
@@ -24,13 +24,13 @@ package gotgbot
 
 	consts.WriteString(generateParseModeConsts())
 
-	stickerTypeConsts, err := generateStickerTypeConsts(d)
+	stickerTypeConsts, err := generateTypeConsts(d, "Sticker")
 	if err != nil {
 		return fmt.Errorf("failed to generate consts for sticker types: %w", err)
 	}
 	consts.WriteString(stickerTypeConsts)
 
-	chatTypeConsts, err := generateChatTypeConsts(d)
+	chatTypeConsts, err := generateTypeConsts(d, "Chat")
 	if err != nil {
 		return fmt.Errorf("failed to generate consts for chat types: %w", err)
 	}
@@ -59,13 +59,13 @@ func generateUpdateTypeConsts(d APIDescription) (string, error) {
 	return out.String(), nil
 }
 
-func generateStickerTypeConsts(d APIDescription) (string, error) {
-	updType, ok := d.Types["Sticker"]
+func generateTypeConsts(d APIDescription, typeName string) (string, error) {
+	updType, ok := d.Types[typeName]
 	if !ok {
-		return "", errors.New("missing 'Sticker' type data")
+		return "", errors.New("missing '" + typeName + "' type data")
 	}
 	out := strings.Builder{}
-	out.WriteString("// The consts listed below represent all the sticker types that can be obtained from telegram.\n")
+	out.WriteString("// The consts listed below represent all the " + strings.ToLower(typeName) + " types that can be obtained from telegram.\n")
 	out.WriteString("const (\n")
 	for _, f := range updType.Fields {
 		if f.Name != "type" {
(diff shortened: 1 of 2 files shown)


scripts/generate/consts.go
```diff
@@ -79,6 +85,32 @@ func generateStickerTypeConsts(d APIDescription) (string, error) {
 	return out.String(), nil
 }
 
+func generateChatTypeConsts(d APIDescription) (string, error) {
```
vassilit: Could/should it be factorised together with with `generateStickerTypeConsts()`, which looks like almost identical to me ?

ALiwoto: I think it's better we keep it as simple as possible and write separated functions for them.
Also, other constant types do have their own functions for auto-generating, so I did what was already being done in the codebase.

PaulSonOfLars: Let's refactor it later, and see if this works first. 

Do you want to regenerate the library so we can see the changes actually appear in the code? Docs are in the README!

ALiwoto: Right, regenerated the code as well

ALiwoto:
Also, @PaulSonOfLars, as for the code refactoring that you and @vassilit mentioned, I can suggest the following function:
```go

func generateTypeConsts(d APIDescription, typeName string) (string, error) {
	updType, ok := d.Types[typeName]
	if !ok {
		return "", errors.New("missing '" + typeName + "' type data")
	}
	out := strings.Builder{}
	out.WriteString("// The consts listed below represent all the " + typeName + " types that can be obtained from telegram.\n")
	out.WriteString("const (\n")
	for _, f := range updType.Fields {
		if f.Name != "type" {
			// the field we want to look at is called "type", ignore all others.
			continue
		}
		types, err := extractQuotedValues(f.Description)
		if err != nil {
			return "", fmt.Errorf("failed to get quoted types: %w", err)
		}
		for _, t := range types {
			constName := typeName + "Type" + snakeToTitle(t)
			out.WriteString(writeConst(constName, t))
		}
	}
	out.WriteString(")\n\n")
	return out.String(), nil
}
```

Then we can use it like this in [here](https://github.com/PaulSonOfLars/gotgbot/blob/371f60d4841ef41c6b99542f3157f97be45ca5c9/scripts/generate/consts.go#L19):
```go
	updateConsts, err := generateTypeConsts(d, "Update")
	if err != nil {
		return fmt.Errorf("failed to generate consts for update types: %w", err)
	}
	consts.WriteString(updateConsts)

	consts.WriteString(generateParseModeConsts())

	stickerTypeConsts, err := generateTypeConsts(d, "Sticker")
	if err != nil {
		return fmt.Errorf("failed to generate consts for sticker types: %w", err)
	}
	consts.WriteString(stickerTypeConsts)

	chatTypeConsts, err := generateTypeConsts(d, "Chat")
	if err != nil {
		return fmt.Errorf("failed to generate consts for chat types: %w", err)
	}
	consts.WriteString(chatTypeConsts)

```

So, instead of defining a separated function for each of them, we can just simply pass the name of the type consts we would like to get, by only defining one single function.

PaulSonOfLars:
@ALiwoto Looks good to me, feel free to refactor! Might be worth lowercasing the comment though, so it doesn't look odd in the comment. Oh, and renaming the variable from `updType` to something like `typeData` or similar, since its not just `Update`. (that one is entirely my fault, i didnt change it when i added the sticker func!)

```
out.WriteString("// The consts listed below represent all the " + strings.ToLower(typeName) + " types that can be obtained from telegram.\n")
```


And by the way - I tihnk your refactor is invalid for the `Update`type. That check just iterates over all update fields, it DOESNT look at the "type" field like sticker and chat do.

ALiwoto:
> @ALiwoto Looks good to me, feel free to refactor! Might be worth lowercasing the comment though, so it doesn't look odd in the comment. Oh, and renaming the variable from `updType` to something like `typeData` or similar, since its not just `Update`. (that one is entirely my fault, i didnt change it when i added the sticker func!)
> 
> ```
> out.WriteString("// The consts listed below represent all the " + strings.ToLower(typeName) + " types that can be obtained from telegram.\n")
> ```
> 
> And by the way - I tihnk your refactor is invalid for the `Update`type. That check just iterates over all update fields, it DOESNT look at the "type" field like sticker and chat do.

Yes, I just tested that, doesn't work for `Update`, sadly.

ALiwoto:
There is also another way around for that, but it will kinda make code ugly (and a bit complicated to read), so I think we just better have separated functions for `Update` consts.

PaulSonOfLars (approved):
Nice one, LGTM! :)

PR #81, opened 2023-01-13
Use go mod edit for ensuring sample bot go mod files

PaulSonOfLars:
# What

Replace sed usage with `go mod edit ...` commands to reduce script brittle-ness.

# Impact

### DIFF
--- a/scripts/ci/ensure-consistent-sample-mod-files.sh
+++ b/scripts/ci/ensure-consistent-sample-mod-files.sh
@@ -6,13 +6,7 @@ SAMPLES_DIR="samples"
 REPO="github.com/PaulSonOfLars/gotgbot" # Current library import path
 GO_VERSION="1.15"                       # Go version we expect our samples to be using
 V_MAJOR="v2"                            # Current major version for the library
-
-# We need to do some bash magic for systems that aren't using gnused
-sedCmd="sed"
-if [[ "$(uname)" == "Darwin" ]]; then
-  command -v gsed || (echo "gnu-sed not installed. Please install with 'brew install gnu-sed'" && return 1)
-  sedCmd="gsed"
-fi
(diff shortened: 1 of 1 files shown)


scripts/ci/ensure-consistent-sample-mod-files.sh
```diff
@@ -6,27 +6,31 @@ SAMPLES_DIR="samples"
...
-
-# We need to do some bash magic for systems that aren't using gnused
-sedCmd="sed"
-if [[ "$(uname)" == "Darwin" ]]; then
-  command -v gsed || (echo "gnu-sed not installed. Please install with 'brew install gnu-sed'" && return 1)
-  sedCmd="gsed"
-fi
+V_DUMMY="v2.99.99"                      # dummy version for the library
 
 for d in "${SAMPLES_DIR}"/*; do
```
vassilit, line 11:
```suggestion
for d in "${SAMPLES_DIR}"/*/; do
```

vassilit (changes requested):

scripts/ci/ensure-consistent-sample-mod-files.sh
vassilit: you can remove the `-d` check, because it lists only directories. But you get a '/' at the end of the `${d}` variable that you need to address as `${d%/}` if you want it removed :

`goModFile="${d}go.mod"`

but 

```
  go mod edit \
    -module "${REPO}/${d%/}" \
```

PaulSonOfLars: I write bash scripts for them to be maintained by regular people, not to show off my bash wizardry.

I would rather have a script which has clear comments explaining we only want to check directories, rather than having to do bash string manipulation and have things like `goModFile="${d}go.mod"` (which is quite unreadable)

PR #83, opened 2023-02-28
update

abhint:
https://github.com/apache/trafficcontrol/issues/6661


PaulSonOfLars:
The GitHub issue you're linking is irrelevant to this library.

>"Go get cannot be used outside a module"

Correct - this is a library, it should only be used in the context of go modules.

> "Go get no longer installs a binary"

Correct - this is a library, not a binary; there is nothing to execute.


If you don't have a go module set up, then you should make one - but that is an issue with your environment, not the library.

PR #98, opened 2023-07-30
Make the Updater's multibot state thread safe

PaulSonOfLars:
# What

This PR makes it possible to add/remove bots to the `Updater` on the fly, rather than the previous option which only allowed for adding bots, without being able to remove them.

# Impact

### DIFF
--- a/ext/updater.go
+++ b/ext/updater.go
@@ -164,7 +164,10 @@ func (u *Updater) StartPolling(b *gotgbot.Bot, opts *PollingOpts) error {
 	updateChan := make(chan json.RawMessage)
 	pollChan := make(chan struct{})
 
-	u.botMapping.addBot(b, updateChan, pollChan, "")
+	err := u.botMapping.addBot(b, updateChan, pollChan, "")
+	if err != nil {
+		return fmt.Errorf("failed to add bot with long polling: %w", err)
+	}
 
 	go u.Dispatcher.Start(b, updateChan)
 	go u.pollingLoop(b, reqOpts, pollChan, updateChan, v)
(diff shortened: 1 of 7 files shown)


ext/updater.go
```diff
@@ -287,6 +259,33 @@ func (u *Updater) Stop() error {
...
+	return true
+}
+
+func (u *Updater) StopAllBots() {
+	for _, bData := range u.botMapping.removeAllBots() {
+		stopBot(bData)
+	}
+}
+
+func stopBot(data botData) {
```
rusq: Conceptually, this could be a method on botData?
```suggestion
func (data botData) stop() {
```

rusq: Also, probably it would make sense to have "stopped atomic.Bool" on the `botData` structure, and write "true" into it on "stop()".  This would prevent panic, should stop be called more than once (by mistake).

rusq (approved):

samples/echoMultiBot/main.go
```diff
@@ -130,7 +130,9 @@ func startWebhookBots(updater *ext.Updater, bots []*gotgbot.Bot, domain string,
 
 	// We add all the bots to the updater.
 	for idx, b := range bots {
-		updater.AddWebhook(b, b.GetToken(), opts)
+		err = updater.AddWebhook(b, b.GetToken(), opts)
+		return fmt.Errorf("failed to add bot: %w", err)
```
rusq: If my eyes do not betray me, this will terminate on the first iteration with:
"failed to add bot: \<nil\>"
Shouldn't this be 
```suggestion
		if err !=nil { return fmt.Errorf("failed to add bot: %w", err) }
```

?

rusq (commented):
@PaulSonOfLars one comment, for your consideration please

samples/echoMultiBot/main.go
PaulSonOfLars: Yes, good catch! Thanks!

PR #103, opened 2023-08-20
Improve flexibility for the dispatcher, and provide an example on adding metrics 

PaulSonOfLars:
# What

Multiple changes in this PR, namely:
- extend and improves the Dispatcher to allow for a way to change or influence the update processing logic. This allows for a simple way to plug in various metrics and monitoring solutions.
- add new example bots that make use of these new dispatcher features (and demonstrate how to add monitoring to a bot!)
- rework the Bot/BaseBotClient structs, as well as the BotClient interface, to reduce coupling and move the token to a more logical place. The client should not be handling auth, just the connection itself. 
- Remove the overly-complex middleware logic for the bot struct; this can be replaced with a custom bot client instead, which avoid ordering issues and is easier to understand without needing weird anonymous wrapper functions.

# Impact

- Are your changes backwards compatible? No - multiple lib changes


samples/metricsBot/metricsMiddleware.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/PaulSonOfLars/gotgbot/v2"
+)
+
+var (
+	totalRequests = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "gotgbot",
+			Name:      "requests_total",
+			Help:      "Number of requests made to the bot API.",
```
roj1512, line 21: Made to or received from the Bot API?

roj1512, line 21:
```suggestion
			Help:      "Number of requests made to the Bot API.",
```

samples/metricsBot/metricsMiddleware.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+		[]string{
+			"api_method",
+		},
+	)
+	totalAPIErrors = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "gotgbot",
+			Name:      "api_request_errors_total",
+			Help:      "Number of bot API errors obtained.",
```
roj1512, line 41:
```suggestion
			Help:      "Number of Bot API errors obtained.",
```

samples/metricsBot/metricsMiddleware.go
```diff
@@ -0,0 +1,101 @@
...
+			"api_method",
+			"api_status_code",
+			"description",
+		},
+	)
+	requestDuration = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "gotgbot",
+			Name:      "api_request_time_seconds",
+			Help:      "Duration of requests made to the bot API.",
```
roj1512, line 53:
```suggestion
			Help:      "Duration of requests made to the Bot API.",
```

samples/metricsBot/metricsMiddleware.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+		[]string{
+			"api_method",
```
itsLuuke: http_method label is missing in this metric

itsLuuke (approved):

samples/metricsBot/metricsMiddleware.go
PaulSonOfLars: I removed the http_method label because the bot client only uses POST methods; so the field ends up being redundant. 

Additionally, that's not information that gets exposed in the error returned, so I figured it would be best to remove it rather than hardcode it

PR #141, opened 2024-02-15
Update callbackquery.go

bcncalling:
- [ ] Added error logging in HandleUpdate function.
- [ ] Refactored Name method to generate a more unique name.

### DIFF
--- a/ext/handlers/callbackquery.go
+++ b/ext/handlers/callbackquery.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"fmt"
+	"log"
 
 	"github.com/PaulSonOfLars/gotgbot/v2"
 	"github.com/PaulSonOfLars/gotgbot/v2/ext"
@@ -11,27 +12,31 @@ import (
 type CallbackQuery struct {
 	AllowChannel bool
 	Filter       filters.CallbackQuery
-	Response     Response
+	Response     func(b *gotgbot.Bot, ctx *ext.Context) error
 }
 
-func NewCallback(filter filters.CallbackQuery, r Response) CallbackQuery {
-	return CallbackQuery{
+func NewCallback(filter filters.CallbackQuery, response func(b *gotgbot.Bot, ctx *ext.Context) error) *CallbackQuery {
+	return &CallbackQuery{
 		Filter:   filter,
-		Response: r,
+		Response: response,
 	}
 }
 
-// SetAllowChannel Enables channel messages for this handler.
-func (cb CallbackQuery) SetAllowChannel(allow bool) CallbackQuery {
+// SetAllowChannel enables channel messages for this handler.
+func (cb *CallbackQuery) SetAllowChannel(allow bool) *CallbackQuery {
 	cb.AllowChannel = allow
 	return cb
 }
 
-func (cb CallbackQuery) HandleUpdate(b *gotgbot.Bot, ctx *ext.Context) error {
-	return cb.Response(b, ctx)
+func (cb *CallbackQuery) HandleUpdate(b *gotgbot.Bot, ctx *ext.Context) error {
+	err := cb.Response(b, ctx)
+	if err != nil {
+		log.Printf("Error handling callback query: %v", err)
+	}
+	return err
 }
 
-func (cb CallbackQuery) CheckUpdate(b *gotgbot.Bot, ctx *ext.Context) bool {
+func (cb *CallbackQuery) CheckUpdate(b *gotgbot.Bot, ctx *ext.Context) bool {
 	if ctx.CallbackQuery == nil {
 		return false
 	}
@@ -43,6 +48,6 @@ func (cb CallbackQuery) CheckUpdate(b *gotgbot.Bot, ctx *ext.Context) bool {
 	return cb.Filter == nil || cb.Filter(ctx.CallbackQuery)
 }
 
-func (cb CallbackQuery) Name() string {
-	return fmt.Sprintf("inlinequery_%p", cb.Response)
+func (cb *CallbackQuery) Name() string {
+	return fmt.Sprintf("callback_query_handler_%p", cb.Response)
 }


ext/handlers/callbackquery.go
```diff
@@ -11,27 +12,31 @@ import (
 type CallbackQuery struct {
 	AllowChannel bool
 	Filter       filters.CallbackQuery
-	Response     Response
```
PaulSonOfLars, line 14: What is the reasoning behind this change?

ext/handlers/callbackquery.go
```diff
@@ -11,27 +12,31 @@ import (
...
+	Response     func(b *gotgbot.Bot, ctx *ext.Context) error
 }
 
-func NewCallback(filter filters.CallbackQuery, r Response) CallbackQuery {
-	return CallbackQuery{
+func NewCallback(filter filters.CallbackQuery, response func(b *gotgbot.Bot, ctx *ext.Context) error) *CallbackQuery {
+	return &CallbackQuery{
 		Filter:   filter,
-		Response: r,
```
PaulSonOfLars, line 20: Please avoid unnecessary changes like this one (especially given there is no explanation provided)

ext/handlers/callbackquery.go
```diff
@@ -11,27 +12,31 @@ import (
...
+		Response: response,
 	}
 }
 
-// SetAllowChannel Enables channel messages for this handler.
-func (cb CallbackQuery) SetAllowChannel(allow bool) CallbackQuery {
+// SetAllowChannel enables channel messages for this handler.
+func (cb *CallbackQuery) SetAllowChannel(allow bool) *CallbackQuery {
```
PaulSonOfLars, line 26: Why are we changing to a pointer reference here?

ext/handlers/callbackquery.go
```diff
@@ -11,27 +12,31 @@ import (
...
 	return cb
 }
 
-func (cb CallbackQuery) HandleUpdate(b *gotgbot.Bot, ctx *ext.Context) error {
-	return cb.Response(b, ctx)
+func (cb *CallbackQuery) HandleUpdate(b *gotgbot.Bot, ctx *ext.Context) error {
+	err := cb.Response(b, ctx)
+	if err != nil {
+		log.Printf("Error handling callback query: %v", err)
+	}
```
PaulSonOfLars, line 35: Why should the callbackquery handler log errors, when none of the others do? 

I'm not a fan of this pattern - I think people should add an error handler to the dispatcher, to ensure all errors are logged properly. Alternatively, they can also wrap the callbackquery handler to log just errors if necessary.

ext/handlers/callbackquery.go
```diff
@@ -43,6 +48,6 @@ func (cb CallbackQuery) CheckUpdate(b *gotgbot.Bot, ctx *ext.Context) bool {
 	return cb.Filter == nil || cb.Filter(ctx.CallbackQuery)
 }
 
-func (cb CallbackQuery) Name() string {
-	return fmt.Sprintf("inlinequery_%p", cb.Response)
+func (cb *CallbackQuery) Name() string {
+	return fmt.Sprintf("callback_query_handler_%p", cb.Response)
```
PaulSonOfLars, line 52: Good catch here, thanks for finding it.

PaulSonOfLars (changes requested):
Hi, thank you for your contribution!

I have to admit I'm not sure what the goal of this PR is - you seem to have opened this up without any prior discussion or context, so I can't tell what you're trying to achieve.

As it stands, the only part that I think should be merged would be the change to the callbackquery name, which you've correctly identified as being wrong. Could you update the PR to only include that change please? Thank you :)

ext/handlers/callbackquery.go
TG-BOTSNETWORK: the signature (b *gotgbot.Bot, ctx *ext.Context) error, allowing for greater versatility in defining response behavior.

ext/handlers/callbackquery.go
TG-BOTSNETWORK: Changing the receiver from CallbackQuery to *CallbackQuery in the SetAllowChannel method allows the method to modify the CallbackQuery struct directly, rather than working with a copy of the struct.

ext/handlers/callbackquery.go
TG-BOTSNETWORK: It's indeed a matter of design preference and code organization. Logging errors within the callback query handler can clutter the handler's logic and may not be the best approach for all situations.

ext/handlers/callbackquery.go
TG-BOTSNETWORK: Ok i will pr this seperate by in the evening due to iam in college now

ext/handlers/callbackquery.go
PaulSonOfLars: That is the same signature as the existing `Response` type... This is a noop and completely unnecessary change.
The definition is already: `type Response func(b *gotgbot.Bot, ctx *ext.Context) error`

ext/handlers/callbackquery.go
PaulSonOfLars: I understand the function of the change, but this doesnt explain the why.

Currently, SetAllowChannel copies the receiver and returns it, allowing for immutable chaining of methods. Changing it to be a pointer would break the immutability and introduce side effects to the method, which would then be inconsistent with all other handlers.

ext/handlers/callbackquery.go
PaulSonOfLars: This is not a matter of "preference", it is a matter of consistency and configurability.

This change introduces a new logging paradigm which is inconsistent with all other handlers. Additionally, it cannot be configured (eg, to use JSON logging over text logging, to silence certain errors but allow others, or even to assign warn/error levels). All of these things are already possible when using the dispatcher-level error handling logic.
 
Ultimately, these are large changes which should be discussed over an issue, or in the support group, before being opened in a PR without any context. I don't like having to close PRs that people have worked on without discussion, because I hate to see work go to waste - so please make sure to discuss these things first.

PaulSonOfLars:
Closing due to being superseded by #144

PR #142, opened 2024-02-15
Add reactions support.

ti-bone:
# What

This introduces support for handling new reactions that are set on messages.
Adds new filters:
`Reaction.FromUserID(id int64)` - Checks whether the reaction from specified user or not.
`Reaction.FromAnonymousChatID(id int64)` - Checks whether the reaction from specified anonymous channel\admin or not.
`Reaction.FromChatID(id int64)` - Checks if the reaction has been set in specified chat
`Reaction.New(Old)ReactionIn(reaction string)` - Checks if the newly set(removed) reaction contains emoji, which is provided as `reaction` argument.

# Impact

- Have you included documentation, or updated existing documentation? Not required.
- Do errors and log messages provide enough context? Not required.

### DIFF
--- a/ext/handlers/filters/reaction/reaction.go
+++ b/ext/handlers/filters/reaction/reaction.go
@@ -9,18 +9,12 @@ func All(_ *gotgbot.MessageReactionUpdated) bool {
 	return true
 }
 
-func FromUserID(id int64) filters.Reaction {
+func FromPeer(id int64) filters.Reaction {
 	return func(mru *gotgbot.MessageReactionUpdated) bool {
 		if mru.User != nil {
 			return mru.User.Id == id
 		}
 
-		return false
-	}
-}
-
-func FromAnonymousChatID(id int64) filters.Reaction {
-	return func(mru *gotgbot.MessageReactionUpdated) bool {
 		if mru.ActorChat != nil {
 			return mru.ActorChat.Id == id
 		}
@@ -29,13 +23,13 @@ func FromAnonymousChatID(id int64) filters.Reaction {
 	}
 }
 
-func FromChatID(id int64) filters.Reaction {
+func ChatID(id int64) filters.Reaction {
 	return func(mru *gotgbot.MessageReactionUpdated) bool {
 		return mru.Chat.Id == id
 	}
 }
 
-func NewReactionIn(reaction string) filters.Reaction {
+func NewReactionEmoji(reaction string) filters.Reaction {
 	return func(mru *gotgbot.MessageReactionUpdated) bool {
 		for _, r := range mru.NewReaction {
 			if r.MergeReactionType().Emoji == reaction {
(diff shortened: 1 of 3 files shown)


ext/handlers/filters/reaction/reaction.go
```diff
@@ -0,0 +1,60 @@
...
+	return func(mru *gotgbot.MessageReactionUpdated) bool {
+		if mru.ActorChat != nil {
+			return mru.ActorChat.Id == id
+		}
+
+		return false
+	}
+}
+
+func FromChatID(id int64) filters.Reaction {
```
PaulSonOfLars: Could we rename this to just be ChatID? I think that would match better with other handlers; its not "from" the chat, its "in" the chat

ext/handlers/filters/reaction/reaction.go
```diff
@@ -0,0 +1,60 @@
...
+	return func(mru *gotgbot.MessageReactionUpdated) bool {
+		return mru.Chat.Id == id
+	}
+}
+
+func NewReactionIn(reaction string) filters.Reaction {
```
PaulSonOfLars: Can we rename this (and the next one) to New(Old)ReactionEmoji? Makes sense to me since that's what we're looking at

PaulSonOfLars (commented):
Looks good to me! Some small changes but otherwise very happy with this. 
Thanks a lot!

ext/handlers/filters/reaction/reaction.go
```diff
@@ -0,0 +1,60 @@
...
+	return func(mru *gotgbot.MessageReactionUpdated) bool {
+		if mru.User != nil {
+			return mru.User.Id == id
+		}
+
+		return false
+	}
+}
+
+func FromAnonymousChatID(id int64) filters.Reaction {
```
PaulSonOfLars: Random thought: should we be merging this to a single item? How are the other handlers doing it? I'm not sure it makes sense to have both, since only one can trigger.

ti-bone: Hm, yep, correct.

ext/handlers/filters/reaction/reaction.go
ti-bone: Sure, looks better.

ext/handlers/filters/reaction/reaction.go
ti-bone: Yep, same here.

ext/handlers/filters/reaction/reaction.go
ti-bone: How do you think about merge them and name as `FromPeer`? 
I don't know how Telegram names such things exactly in BotAPI, but in MTProto it's named usually as `Peer`.

ti-bone:
`Reaction.FromPeer(id int64)` - Was renamed and merged from `FromUserID` and `FromAnonymousChatID`, checks if the specified peer(user or anonymous channel) sent this reaction update.

PR #159, opened 2024-05-09
[Draft] Add new message.CommandName filters

PaulSonOfLars:
# What

This change adds a few new message filters to allow `MessageHandlers` to function similarly to `CommandHandlers`. With this new change, a `CommandHandler` can be considered a specialised `MessageHandler` (with command filters pre-applied).

Worth noting that with this change, the `CommandHandler` type becomes significantly less useful; maybe the type should be removed, and the NewCommand method should be left simply as a wrapper around NewMessage?

# Impact

- Are your changes backwards compatible? Yes; unless we choose to remove the CommandHandler type.

### DIFF
--- a/ext/handlers/command.go
+++ b/ext/handlers/command.go
@@ -2,10 +2,10 @@ package handlers
 
 import (
 	"strings"
-	"unicode/utf8"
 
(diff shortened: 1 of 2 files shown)


PaulSonOfLars:
Closing due to no interest from community

PR #179, opened 2024-07-12
Add constants for chat actions

megum1n:
# What

Add constants for chat actions to avoid typos.
https://core.telegram.org/bots/api#sendchataction

# Impact

- Have you included documentation, or updated existing documentation? N/A

### DIFF
--- a/scripts/generate/consts.go
+++ b/scripts/generate/consts.go
@@ -3,6 +3,7 @@ package main
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"strings"
 )
 
@@ -23,7 +24,12 @@ package gotgbot
 	consts.WriteString(updateConsts)
 
 	consts.WriteString(generateParseModeConsts())
-	consts.WriteString(generateChatActionConsts())
+
+	chatActions, err := generateChatActionConsts(d)
+	if err != nil {
+		return fmt.Errorf("failed to generate consts for chat actions: %w", err)
+	}
+	consts.WriteString(chatActions)
 
 	stickerTypeConsts, err := generateTypeConsts(d, "Sticker")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


scripts/generate/consts.go
```diff
@@ -122,6 +123,35 @@ func generateParseModeConsts() string {
...
+		"record_video",
+		"upload_video",
+		"record_voice",
+		"upload_voice",
+		"upload_document",
+		"choose_sticker",
+		"find_location",
+		"record_video_note",
+		"upload_video_note",
+	}
```
PaulSonOfLars: Would it be possible for this list to be obtained from the type spec?
For example, by using some regexes on the sendChatAction method's "action" field description? 

If the list is defined manually, then it might as well not be generated; it's just a manually written list, and any changes on telegrams side wouldn't reflect.
I know it's done this way for formatting consts, but that was intended as an exception; parsemodes aren't expected to change, and are relatively short list!

PaulSonOfLars (commented):
Hi, thank you for your contribution! Can definitely see the value in this, so appreciate the input :) 

One issue though; this isn't generating from the spec. Would it be possible to change that? Thanks in advance!

scripts/generate/consts.go
megum1n: @PaulSonOfLars I agree, let's try to actually generate it. 
I decided not to create universal function in this case because it would require a lot of input.
```go
func generateConstsFromMethodFieldDescription(d APIDescription, methodName string, fieldName string, regexp string) (string, error) {
    ...
}
```

Please let me know if current code sucks.

PaulSonOfLars (approved):
Awesome stuff, exactly what I was thinking of!

PR #182, opened 2024-07-18
Draft: Add reaction constants

megum1n:
# What

Just like in #179, it would be great to have constants for all available reactions. 
I've added [new dependency](https://github.com/forPelevin/gomoji) to get reaction name from the emoji string. 

### Alternatives
- Find JSON with all emojis somewhere on github, which has a map with emoji labels. Scrub trough it to find names for emojis
- Generate such file ourselves via scrape script in https://github.com/PaulSonOfLars/telegram-bot-api-spec
- There might be an easier way to do that w/o additional dependencies, but I couldn't find it 😓 

# Impact

### DIFF
--- a/gen_consts.go
+++ b/gen_consts.go
@@ -111,6 +111,21 @@ const (
 	ParseModeNone       = ""
 )
 
+// The consts listed below represent all the sticker types that can be obtained from telegram.
+const (
+	StickerTypeRegular     = "regular"
+	StickerTypeMask        = "mask"
+	StickerTypeCustomEmoji = "custom_emoji"
+)
+
+// The consts listed below represent all the chat types that can be obtained from telegram.
+const (
+	ChatTypePrivate    = "private"
+	ChatTypeGroup      = "group"
+	ChatTypeSupergroup = "supergroup"
+	ChatTypeChannel    = "channel"
+)
+
 // The consts listed below represent all the chat action options that can be sent to telegram.
 const (
 	ChatActionTyping          = "typing"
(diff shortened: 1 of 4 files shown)


annihilatorrrr:
Maybe: https://github.com/PaulSonOfLars/gotgbot#features
> No third party library bloat; only uses standard library.

megum1n:
@annihilatorrrr 
> No third party library bloat; only uses standard library.

gotgbot package itself still won't use any bloat. Only script to generate code will use external package.

PaulSonOfLars:
Hey, thank you for your patience on my reply.

I know what you mean by thi not adding dependencies in practice, but it is in the root go.mod file, so there is a questionmark there. Would be best if they weren't there...

One option would be to make the generation files their own module; could we try creating some new go.mod and go.sum files for the codegen only? That should allow us to separate things as needed, and keep the lib dependency-free :)

liuerfire:
How about make another repo for plugins or handlers need extra deps like [python-telegram-bot](https://github.com/python-telegram-bot/python-telegram-bot) has [ptbcontrib](https://github.com/python-telegram-bot/ptbcontrib)?

PaulSonOfLars:
@liuerfire That is indeed something I've considered. 

Happy to open one up if theres demand :)
What kind of things would people be looking for?
- emoji constants
- logging tools?
- Rate limit wrappers? That kind of thing?

liuerfire:
the most I want is persistent states for conversation :-)

PR #183, opened 2024-07-22
Check for command entity first in Command.checkMessage

Jisin0:
# What
Minor improvement on Command.checkMessage to look for a command entity before trying to parse the command.

# Impact

- Are your changes backwards compatible? **Yes**
- Have you included documentation, or updated existing documentation? **N/A**
- Do errors and log messages provide enough context? **N/A**


PaulSonOfLars:
Hi, thanks for raising!

I understand what the PR does, but it's not clear to me _why_ it's necessary; is the existing logic causing issues? Are there any errors to be shared?

Jisin0:
There's no issue with the current setup. But it made more sense to check if there is a command entity before splitting and parsing the command, which I assume will do for all messages unnecessarily that reach the handler.

PaulSonOfLars (approved):
Fair enough, hadn't thought about it that way. Good catch!

In the future, please add that to the PR description so there's some context to your changes :)

Issue #190, opened 2024-08-21
Add context.Context to the func signatures in gen_*.go


liuerfire:
For backwards compatibility, maybe add extra methods like 
```go
func (bot *Bot) SendMessageWithContext(ctx context.Context, chatId int64, text string, opts *SendMessageOpts) (*Message, error)
```

## Why need this 

For example, we have a web server, then need to send messages to users if they make some actions on our website. A typical code maybe like this:

```go
func (api *API) handleAction(w http.ResponseWriter, r *http.Request) {
    ctx := r.Context()
    // do something with the request's context, like querying the db
    db.getUser(ctx, userid)
   
    // Everything is done and need to notify the user. 
    // However we lost the context for something like tracing, logging or timeout control 
    // since SendMessage doesn't accept `Context`
    bot.SendMessage(chatId int64, text string, opts *SendMessageOpts)
}
```

Hope I made my point clear 😄

PaulSonOfLars:
Hey hey! thanks for opening. 
Yes, of course - that makes complete sense. Was wondering when someone would ask for this to be honest 😅 

I'd be happy to to add <x>WithContext (or even <x>WithCtx if thats preferred? not sure what the go ecosystem trends to). 
Just need to come up with a clean(ish) way of representing that without blindly duplicating the code; shouldn't be too difficult!

PaulSonOfLars:
Let me know how #192 looks to you!

I'd really appreciate it if you could try out the branch on your side, and run any tests you might need - this would be a really helpful datapoint so I can make sure everything is working as expected (and is as intuitive as possible!)

PR #192, opened 2024-08-26
Add WithContext methods

PaulSonOfLars:
# What

Add *WithContext methods when dealing with the telegram API, to ensure that we can set the right deadlines where necessary and tell any long-running methods when they should be cancelled.

Addresses #190 

# Impact

- Are your changes backwards compatible? No - some small changes to the BotClient interface; `TimeoutContext` now takes an optional parent context too.


fishronsage:
LGTM. My code is running good with this branch.

liuerfire:
LGTM. My code is running good with this branch.

request.go
```diff
@@ -22,8 +22,8 @@ const (
 type BotClient interface {
 	// RequestWithContext submits a POST HTTP request a bot API instance.
 	RequestWithContext(ctx context.Context, token string, method string, params map[string]string, data map[string]FileReader, opts *RequestOpts) (json.RawMessage, error)
-	// TimeoutContext calculates the required timeout contect required given the passed RequestOpts, and any default opts defined by the BotClient.
-	TimeoutContext(opts *RequestOpts) (context.Context, context.CancelFunc)
+	// TimeoutContext wraps the existing context with the timeout from the provided RequestOpts, or any default opts defined by the BotClient.
+	TimeoutContext(ctx context.Context, opts *RequestOpts) (context.Context, context.CancelFunc)
```
liuerfire: BTW, just curious about this method 😄 

What's your consideration to define this method separately?
I think we can pass timeout parameter to https://github.com/PaulSonOfLars/gotgbot/blob/v2.0.0-rc.29/bot.go#L42 to achieve the same effect and even if we need ctx to do other things we can do it in `RequestWithContext`.

PaulSonOfLars: Well, two reasons: 
1) bot-wide timeouts
2) backwards compatibility 

I do agree it would be nice if it wasn't part of it at this point! Might look into it to see if it still needs to be part of the interface though...

PaulSonOfLars:
> 
> LGTM. My code is running good with this branch.
> 
> 

Awesome, great to hear - thank you!

request.go
liuerfire: Got. thanks for replying

PaulSonOfLars: Managed to clean it up - you were right, doesn't need to be there after all :)

PR #201, opened 2024-11-01
Accept context when stopping updater

myxzlpltk:
Breaking changes. Allow user to pass context when stop updater


PaulSonOfLars (commented):
Hi @myxzlpltk, thank you for the PR!

All of the methods in this lib have two versions; one without, and one **with** the context (suffixed with `WithContext`). This avoids us having to add breaking changes where they arent strictly necessary

Would it be possible for you to do the same thing for this PR, and add a StopWithContext method to support this? This will also have the advantage of fixing all the breaking tests that test the Stop() functionality :)

PR #202, opened 2024-11-03
Try out using the new log/slog library for layered logging information

PaulSonOfLars:
# What

A few breaking changes to use the new `slog` packages. This will ultimately allow for customising the output format of any log output, thus integrating with existing solutions (and allowing for text/json/custom output), and potentially adding more debug information to the dispatcher logic

# Impact

- Are your changes backwards compatible? No


ext/dispatcher.go
```diff
@@ -127,7 +128,7 @@ func NewDispatcher(opts *DispatcherOpts) *Dispatcher {
 	var errHandler DispatcherErrorHandler
 	var panicHandler DispatcherPanicHandler
 	var unhandledErrFunc ErrorFunc
-	var errLog *log.Logger
+	logger := slog.New(slog.NewTextHandler(os.Stderr, nil))
```
rusq: I think it could be beneficial to use the `slog.Default()` here, so that if the caller had initialised the slog.Default logger before calling NewDispatcher, would be automatically propagated.

ext/updater.go
```diff
@@ -51,42 +52,34 @@ type Updater struct {
...
 	// If nil, logging is done via the log package's standard logger.
-	ErrorLog *log.Logger
+	Logger *slog.Logger
 }
 
 // NewUpdater Creates a new Updater, as well as a Dispatcher and any optional updater configurations (via UpdaterOpts).
 func NewUpdater(dispatcher UpdateDispatcher, opts *UpdaterOpts) *Updater {
 	var unhandledErrFunc ErrorFunc
-	var errLog *log.Logger
+	logger := slog.New(slog.NewTextHandler(os.Stderr, nil))
```
rusq: same comment as above

rusq (approved):

ext/dispatcher.go
PaulSonOfLars: Great catch, thank you! Had forgotten about that entirely :)

ext/dispatcher.go
```diff
@@ -143,7 +141,7 @@ func NewDispatcher(opts *DispatcherOpts) *Dispatcher {
 		errHandler = opts.Error
 		panicHandler = opts.Panic
 		unhandledErrFunc = opts.UnhandledErrFunc
-		errLog = opts.ErrorLog
+		logger = opts.Logger
```
rusq: This is opinionated, but I find that having this type of helper function may help in some cases:
```go
func iftrue[T any](cond bool, t T, f T) T {
   if cond {
      return t
   }
   return f
}
```
(addressing the lack of  C-style ternary condition operator ?: in Go)

and then:
```go
logger = iftrue(opts.Logger == nil, logger, opts.Logger)
```

PaulSonOfLars: Ooh, I like that, thank you!

PR #204, opened 2024-11-11
[Draft] Add custom types for string constants

fullpipe:
# What

I'd like to give string constants their own types. 
So, for RFC, I added only `ParseMode`. 
What do you think?

# Impact

Impact is not that much. But it is easy to read and write.

yes

I would update it if the change is accepted.

### DIFF
--- a/gen_consts.go
+++ b/gen_consts.go
@@ -108,11 +108,13 @@ func (u Update) GetType() string {
 }
 
 // The consts listed below represent all the parse_mode options that can be sent to telegram.
+type ParseMode string
+
 const (
-	ParseModeHTML       = "HTML"
-	ParseModeMarkdownV2 = "MarkdownV2"
-	ParseModeMarkdown   = "Markdown"
-	ParseModeNone       = ""
+	ParseModeHTML       ParseMode = "HTML"
+	ParseModeMarkdownV2 ParseMode = "MarkdownV2"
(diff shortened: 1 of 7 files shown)


PaulSonOfLars:
Hi, thank you for your patience!

Sadly, I don't think this would be a helpful PR to merge - while it's great to have custom types, go's interpretation of `type ParseMode string` means that any string can be treated as the `ParseMode` type.

The meaning of this is the following:

```golang
// This is the intended usage following the new parsemode syntax, and will work fine. Great!
&gotgbot.SendMessageOpts{ParseMode: gotgbot.ParseModeHTML}

// But so will this, because 'string' can be implicitly cast to gotgbot.ParseMode. Less great.
&gotgbot.SendMessageOpts{ParseMode: "HTML"}

// Which then means that this is also valid. 
// This is the exact same as the current state, meaning a ParseMode type doesn't add anything to the lib.
&gotgbot.SendMessageOpts{ParseMode: "nonsense-string"}
```

The end result being that the `ParseMode` type actually doesn't bring any value to the codebase - except by increasing the codegen complexity. 

To be clear - I don't think this is a failure on this PR; on the contrary, I really appreciate the approach and can sympathise with the goals. Ultimately the issue here is to do with Go's lack of enums!
Does that make sense? :)

fullpipe:
Ok, no problem)

PR #210, opened 2024-12-21
Crude implementation of context propagation to Updater

rusq:
# What

Propagate parent Context to updater to prevent it idling till the next update when Stop is called.

This is draft and crude, and is intended to be a starting point for discussion.

# Impact

- A new field is added to the PollingOpts - ParentCtx that allows passing the parent Context.
   - If it is nil, it is initialised with context.Background() to prevent nil pointer dereference.
- Each Request originating from the poller is created with the ParentCtx, when it's cancelled, user may
  observe "Request Failed: context cancelled" in the logs.
- I had to pull the <-bData.stopUpdates to be on the same level as ctx.Done() check, and I don't like it.
  - This happened because I noticed it is called in the handler, I could have used r.Context() instead,
    but wasn't sure of consequences.  What if the request is terminated and that context is dead?

### DIFF
--- a/ext/botmapping.go
+++ b/ext/botmapping.go
@@ -47,8 +47,10 @@ type botMapping struct {
 	errorLog *log.Logger
 }
 
-var ErrBotAlreadyExists = errors.New("bot already exists in bot mapping")
-var ErrBotUrlPathAlreadyExists = errors.New("url path already exists in bot mapping")
+var (
+	ErrBotAlreadyExists        = errors.New("bot already exists in bot mapping")
+	ErrBotUrlPathAlreadyExists = errors.New("url path already exists in bot mapping")
+)
 
 // addBot Adds a new bot to the botMapping structure.
 // Pass an empty urlPath/webhookSecret if using polling instead of webhooks.
--- a/ext/updater.go
+++ b/ext/updater.go
@@ -107,6 +107,8 @@ type PollingOpts struct {
 	//    long-polling, Telegram responds to your request as soon as new messages are available.
 	//    When setting this, it is recommended you set your PollingOpts.Timeout value to be slightly bigger (eg, +1).
 	GetUpdatesOpts *gotgbot.GetUpdatesOpts
+	// ParentCtx is the parent context for the polling loop.
+	ParentCtx context.Context
 }
 
 // StartPolling starts polling updates from telegram using getUpdates long-polling.
@@ -155,6 +157,9 @@ func (u *Updater) StartPolling(b *gotgbot.Bot, opts *PollingOpts) error {
 				v["allowed_updates"] = string(bs)
 			}
 		}
+		if opts.ParentCtx == nil {
+			opts.ParentCtx = context.Background()
+		}
 	}
 
 	bData, err := u.botMapping.addBot(b, "", "")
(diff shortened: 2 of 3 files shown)


ext/botmapping.go
```diff
@@ -47,8 +47,10 @@ type botMapping struct {
 	errorLog *log.Logger
 }
 
-var ErrBotAlreadyExists = errors.New("bot already exists in bot mapping")
-var ErrBotUrlPathAlreadyExists = errors.New("url path already exists in bot mapping")
+var (
```
rusq, line 50: Excuse my formatter

ext/updater.go
```diff
@@ -163,24 +168,31 @@ func (u *Updater) StartPolling(b *gotgbot.Bot, opts *PollingOpts) error {
...
 }
 
-func (u *Updater) pollingLoop(bData *botData, opts *gotgbot.RequestOpts, v map[string]string) {
+func (u *Updater) pollingLoop(ctx context.Context, bData *botData, opts *gotgbot.RequestOpts, v map[string]string) {
 	bData.updateWriterControl.Add(1)
 	defer bData.updateWriterControl.Done()
 
 	for {
+		select {
+		case <-bData.stopUpdates:
```
rusq, line 182: This duplicates the check in `shouldStopUpdates`

samples/echoBot/main.go
```diff
@@ -45,13 +50,20 @@ func main() {
 		MaxRoutines: ext.DefaultMaxRoutines,
 	})
 	updater := ext.NewUpdater(dispatcher, nil)
+	go func() {
+		// sentinel, when the context is done, stop all bots.
+		<-ctx.Done()
+		updater.Stop()
+		updater.StopAllBots()
```
rusq, line 57: Initially just called StopAllBots, but that wasn't enough to stop updater from Idle()'ing.

ext/updater.go
```diff
@@ -163,24 +168,31 @@ func (u *Updater) StartPolling(b *gotgbot.Bot, opts *PollingOpts) error {
 	}
 
 	go u.Dispatcher.Start(b, bData.updateChan)
-	go u.pollingLoop(bData, reqOpts, v)
+	go u.pollingLoop(opts.ParentCtx, bData, reqOpts, v)
```
PaulSonOfLars, line 171: note: this risks a nil pointer if opts is nil :)

ext/updater.go
PaulSonOfLars: I don't think we need this; RequestWithContext would handle the exit for us if necessary

samples/echoBot/main.go
PaulSonOfLars: Hmm. that might be worth discussing in itself!

PaulSonOfLars (commented):
Interesting approach!

Definitely helpful to handle `os` signals, but I think it misses the usecase of a bot being able to "stop" itself, which I expect is a common usecase (eg, see the multibot example). What do you think?

I've given it my shot in #211; let me know what you think!

ext/updater.go
rusq: Shoot!

rusq:
Thanks for your comments!  I think #211 nails it, so I'll close this PR.

PR #211, opened 2024-12-21
updater.StopBot should close polling context

PaulSonOfLars:
# What

This should speed up updater.StopBot significantly by allow closing the context used for polling.

# Impact

### DIFF
--- a/ext/botmapping_test.go
+++ b/ext/botmapping_test.go
@@ -18,7 +18,7 @@ func Test_botMapping(t *testing.T) {
 	t.Run("addBot", func(t *testing.T) {
 		// check that bots can be added fine
 		var err error
-		origBdata, err = bm.addBot(b, "", "")
+		origBdata, err = bm.addPollingBot(b, nil)
 		if err != nil {
 			t.Errorf("expected to be able to add a new bot fine: %s", err.Error())
 			t.FailNow()
@@ -31,7 +31,7 @@ func Test_botMapping(t *testing.T) {
 
 	t.Run("doubleAdd", func(t *testing.T) {
 		// Adding the same bot twice should fail
-		_, err := bm.addBot(b, "", "")
+		_, err := bm.addPollingBot(b, nil)
 		if err == nil {
 			t.Errorf("adding the same bot twice should throw an error")
 			t.FailNow()
@@ -84,7 +84,10 @@ func Test_botData_isUpdateChannelStopped(t *testing.T) {
 		BotClient: &gotgbot.BaseBotClient{},
 	}
 
-	bData, err := bm.addBot(b, "", "")
+	ctxCancelled := false
+	bData, err := bm.addPollingBot(b, func() {
+		ctxCancelled = true
+	})
 	if err != nil {
 		t.Errorf("bot with token %s should not have failed to be added", b.Token)
 		return
@@ -98,5 +101,8 @@ func Test_botData_isUpdateChannelStopped(t *testing.T) {
 	if !bData.shouldStopUpdates() {
 		t.Errorf("bot with token %s should be stopped", b.Token)
 		return
+	}
+	if !ctxCancelled {
+		t.Errorf("bot with token %s should have a cancelled context ", b.Token)
 	}
 }
--- a/ext/updater.go
+++ b/ext/updater.go
@@ -168,8 +168,10 @@ func (u *Updater) StartPolling(b *gotgbot.Bot, opts *PollingOpts) error {
 
 	bData.updateWriterControl.Add(1)
 	go func() {
-		u.pollingLoop(ctx, bData, reqOpts, v)
+		// defer, so it gets called even in case of panics.
 		defer bData.updateWriterControl.Done()
+
+		u.pollingLoop(ctx, bData, reqOpts, v)
 	}()
 
 	return nil
--- a/ext/updater.go
+++ b/ext/updater.go
@@ -165,15 +165,17 @@ func (u *Updater) StartPolling(b *gotgbot.Bot, opts *PollingOpts) error {
 	}
 
 	go u.Dispatcher.Start(b, bData.updateChan)
-	go u.pollingLoop(ctx, bData, reqOpts, v)
+
+	bData.updateWriterControl.Add(1)
+	go func() {
+		u.pollingLoop(ctx, bData, reqOpts, v)
+		defer bData.updateWriterControl.Done()
+	}()
 
 	return nil
 }
 
 func (u *Updater) pollingLoop(ctx context.Context, bData *botData, opts *gotgbot.RequestOpts, v map[string]string) {
-	bData.updateWriterControl.Add(1)
-	defer bData.updateWriterControl.Done()
-
 	for {
 		// Check if updater loop has been terminated.
 		if bData.shouldStopUpdates() {
@@ -266,7 +268,7 @@ func (u *Updater) Stop() error {
 	// Stop the dispatcher from processing any further updates.
 	u.Dispatcher.Stop()
 
-	// Finally, atop idling.
+	// Finally, stop idling.
 	if u.stopIdling != nil {
 		close(u.stopIdling)
 	}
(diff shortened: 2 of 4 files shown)


ext/updater.go
```diff
@@ -157,18 +157,20 @@ func (u *Updater) StartPolling(b *gotgbot.Bot, opts *PollingOpts) error {
 		}
 	}
 
-	bData, err := u.botMapping.addBot(b, "", "")
+	ctx, closeFn := context.WithCancel(context.Background())
```
PaulSonOfLars, line 160: We could reuse the opts logic to pass in an updater from #210 here - might be a good way to handle signals?

Though the signals approach would only close the updater, not the dispatcher; so might be best to handle closure explitly?

rusq: I think this approach is much better than in #210 - the library should not be concerned about any of the OS signals — it just needs to know one thing:  "something wants the updater to stop".  And technically it doesn't need the Context, as it's not going to use it for anything else but only to know when it's cancelled.

Handing specific OS signals and termination is a caller's concern, for example, that's how the echoBot could handle the os.Interrupt signal and tell the bot to terminate — works great on this branch:

```diff
diff --git a/samples/echoBot/main.go b/samples/echoBot/main.go
index bdffc88..2f390b1 100644
--- a/samples/echoBot/main.go
+++ b/samples/echoBot/main.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"net/http"
 	"os"
+	"os/signal"
 	"time"
 
 	"github.com/PaulSonOfLars/gotgbot/v2"
@@ -34,6 +36,8 @@ func main() {
 	if err != nil {
 		panic("failed to create new bot: " + err.Error())
 	}
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
+	defer cancel()
 
 	// Create updater and dispatcher.
 	dispatcher := ext.NewDispatcher(&ext.DispatcherOpts{
@@ -45,6 +49,11 @@ func main() {
 		MaxRoutines: ext.DefaultMaxRoutines,
 	})
 	updater := ext.NewUpdater(dispatcher, nil)
+	go func() {
+		<-ctx.Done()
+		log.Println("shutting down...")
+		updater.Stop()
+	}()
 
 	// Add echo handler to reply to all text messages.
 	dispatcher.AddHandler(handlers.NewMessage(message.Text, echo))
```

ext/updater.go
```diff
@@ -180,8 +182,13 @@ func (u *Updater) pollingLoop(bData *botData, opts *gotgbot.RequestOpts, v map[s
 
 		// Manually craft the getUpdate calls to improve memory management, reduce json parsing overheads, and
 		// unnecessary reallocation of url.Values in the polling loop.
-		r, err := bData.bot.Request("getUpdates", v, nil, opts)
+		r, err := bData.bot.RequestWithContext(ctx, "getUpdates", v, nil, opts)
 		if err != nil {
+			if errors.Is(err, context.Canceled) {
```
rusq, line 191: Beautiful

ext/updater_test.go
```diff
@@ -152,6 +152,54 @@ func concurrentTest(t *testing.T) {
...
+	time.Sleep(time.Millisecond * 500)
+
+	// Should still have lots of time until timeout, so... kill the bots.
+	start := time.Now()
+	u.StopAllBots()
+
+	// If it took longer than 1ms then we know something went wrong; ctx should've stopped immediately.
+	since := time.Since(start)
+	t.Logf("stopping took %dms", since.Milliseconds())
+	if since > time.Millisecond {
```
rusq: Nit: do you think that could cause the test to fail on slower systems?  Maybe giving it a bit of a play (i.e. 5-10ms) would be feasible?

ext/botmapping_test.go
```diff
@@ -84,7 +84,10 @@ func Test_botData_isUpdateChannelStopped(t *testing.T) {
 		BotClient: &gotgbot.BaseBotClient{},
 	}
 
-	bData, err := bm.addBot(b, "", "")
+	ctxCancelled := false
```
rusq, line 87: I was suspecting this test might fail if executed with `-race`, but the different one failed instead.

Note: this possibly does not impact the bot's behaviour in real environment, just a quirk of how the test is accessing the data.

```
[9:15:47] 0:_thirdparty/gotgbot/ext (paul/close-polling-ctx)> go test -race ./...
==================
WARNING: DATA RACE
Write at 0x00c000214098 by goroutine 219:
  runtime.racewrite()
      <autogenerated>:1 +0x1e
  github.com/PaulSonOfLars/gotgbot/v2/ext.(*botData).stop()
      /Users/rustam/wp/_thirdparty/gotgbot/ext/botmapping.go:231 +0xbd
  github.com/PaulSonOfLars/gotgbot/v2/ext.(*Updater).StopAllBots()
      /Users/rustam/wp/_thirdparty/gotgbot/ext/updater.go:288 +0xa4
  github.com/PaulSonOfLars/gotgbot/v2/ext.(*Updater).Stop()
      /Users/rustam/wp/_thirdparty/gotgbot/ext/updater.go:264 +0x10d
  github.com/PaulSonOfLars/gotgbot/v2/ext_test.TestUpdaterSupportsTwoPollingBots()
      /Users/rustam/wp/_thirdparty/gotgbot/ext/updater_test.go:434 +0xa52
  testing.tRunner()
      /usr/local/Cellar/go/1.23.4/libexec/src/testing/testing.go:1690 +0x226
  testing.(*T).Run.gowrap1()
      /usr/local/Cellar/go/1.23.4/libexec/src/testing/testing.go:1743 +0x44

Previous read at 0x00c000214098 by goroutine 222:
  runtime.raceread()
      <autogenerated>:1 +0x1e
  github.com/PaulSonOfLars/gotgbot/v2/ext.(*Updater).pollingLoop()
      /Users/rustam/wp/_thirdparty/gotgbot/ext/updater.go:174 +0x8d
  github.com/PaulSonOfLars/gotgbot/v2/ext.(*Updater).StartPolling.gowrap2()
      /Users/rustam/wp/_thirdparty/gotgbot/ext/updater.go:168 +0x79
<... elided for brevity ...>
--- FAIL: TestUpdaterSupportsTwoPollingBots (0.00s)
    testing.go:1399: race detected during execution of test
```

rusq (approved):

ext/botmapping_test.go
rusq: If you're unable to get it to fail, so running like this helps:
```
go test ./... -race -count=10
```
takes a long time tho to run.

ext/updater_test.go
PaulSonOfLars: Sure, will increase - good shout!

ext/botmapping_test.go
PaulSonOfLars: Interesting - theres definitely something funky going on there, not sure what exactly. Will take a look. Might be a different PR, depending on how long it takes! Thanks for the review and the report <3

ext/updater.go
```diff
@@ -157,21 +157,27 @@ func (u *Updater) StartPolling(b *gotgbot.Bot, opts *PollingOpts) error {
...
+	}()
 
 	return nil
 }
 
-func (u *Updater) pollingLoop(bData *botData, opts *gotgbot.RequestOpts, v map[string]string) {
-	bData.updateWriterControl.Add(1)
-	defer bData.updateWriterControl.Done()
-
+func (u *Updater) pollingLoop(ctx context.Context, bData *botData, opts *gotgbot.RequestOpts, v map[string]string) {
```
PaulSonOfLars, line 180: @rusq This should fix it! An edge case for sure, but I can see this being triggered in live systems which are dynamically adding/removing bots.

TL;DR: Race condition caused by the waitgroup being started in a goroutine, rather than before it.

This meant that when the stars (mis)aligned, calling `updater.StartPolling(...)` followed by `updater.Stop()`, could result in `.Wait()` (from the `.Stop()`) before calling `.Add(1)` (from the goroutine in `.StartPolling`); which then causes the race.

The fix makes sure that the `.Add` is done sequentially, thus will always run before the `.Wait`.

rusq (approved):
LGTM, thank you!

ext/updater.go
rusq: Concurrency bugs are always the hardest to debug, great break down!

PR #232, opened 2025-09-11
cleanup: drop temp var in range loop (Go 1.22+ makes loop vars unique)

acquisitionist:
# What

This PR removes the redundant temp := updData assignment inside a for range loop.
In Go versions prior to 1.22, this pattern was required to avoid loop variable reuse issues when taking the address of the variable or capturing it inside goroutines/closures.

Since Go 1.22, each iteration of a for range loop creates a new loop variable, so the extra copy is unnecessary and can be safely removed.

Reference: https://go.dev/blog/loopvar-preview

# Impact

Yes.
 No documentation changes needed.
Errors and log messages are unaffected; behavior is functionally identical.


PaulSonOfLars (approved):
Thank you for the PR!

PR #242, opened 2025-10-19
Simplify requests code and reduce double-unmarshalling for many fields

PaulSonOfLars:
# What

It should be possible to simplify our existing requests code to something slightly more reasonable, removing the existing `application/json` logic. We also change the RequestParams from `params map[string]string` to `params map[string]any`.

This should help by reducing the number of json.marshal calls, thus hopefully reducing CPU and memory usage.

# Impact

- Are your changes backwards compatible? No. This changes method definitions and types.

### DIFF
--- a/bot.go
+++ b/bot.go
@@ -114,14 +114,14 @@ func (bot *Bot) UseMiddleware(mw func(client BotClient) BotClient) *Bot {
 	return bot
 }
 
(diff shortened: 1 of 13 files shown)


Fedasov (approved):

PaulSonOfLars:
Confirmed working through t.me/MissRose_bot - ok to merge

PR #247, opened 2025-11-19
Improve support for retries when dealing with files

PaulSonOfLars:
# What

Currently, sending a request with a filehandle will read the file and leave the request in a state where it cannot be retried.

By checking if the reader type supports seeking, we can reset the file to the start, which then allows us to retry the file reading.

# Impact


stefanodvx (approved):
tested and working ;)

Issue #259, opened 2026-03-13
InputFileByReader buffers entire file in memory causing high memory usage for large files


danielwaichcesl:
## Description

When using `InputFileByReader` to send files (documents, photos, videos, etc.), the entire file is buffered in memory before being sent to Telegram's API. This causes significant memory spikes, especially for large files.

## Environment

- **gotgbot version**: v2 (latest)
- **Go version**: 1.21+
- **OS**: Linux (Docker container)

## Steps to Reproduce

1. Use `InputFileByReader` to send a file via `SendDocument`, `SendPhoto`, `SendVideo`, etc.
2. Monitor memory usage during the upload
3. Observe memory spike proportional to file size

## Example Code

```go
fileStream, _ := downloadFile(ctx, fileURL) // Returns io.ReadCloser
defer fileStream.Close()

// This buffers the entire file in memory
response, err := bot.SendDocument(
    chatId,
    gotgbot.InputFileByReader("file.pdf", fileStream),
    &gotgbot.SendDocumentOpts{
        Caption: "Test",
    },
)
```

## Expected Behavior

The library should stream the file directly to Telegram's API without buffering the entire content in memory, similar to how standard HTTP multipart uploads work with `io.Copy`.

## Actual Behavior

**Memory usage for a 31 MB file:**
- Before upload: ~6 MB
- During upload: ~66 MB (**+60 MB spike**)
- After upload: ~6 MB (after GC)

The memory spike is approximately **2x the file size**, which becomes problematic for:
- Large files (100+ MB)
- High-concurrency scenarios
- Memory-constrained environments (containers with limited RAM)

## Root Cause Analysis

After investigating the code, it appears that `InputFileByReader` reads the entire stream into a buffer to:
1. Determine the file size for the `Content-Length` header
2. Set multipart boundaries correctly

However, Telegram's Bot API supports chunked transfer encoding and doesn't strictly require `Content-Length` for file uploads.

## Proposed Solution

Add a streaming option that:
1. Uses `io.Pipe` to connect the reader directly to the HTTP request body
2. Writes multipart data on-the-fly without intermediate buffering
3. Allows the HTTP client to handle chunked transfer encoding

## Workaround

Currently, we're working around this by manually constructing the multipart request:

```go
func uploadFileStreaming(ctx context.Context, botToken string, chatId int64, 
    filename string, fileReader io.Reader, endpoint string, fileFieldName string) (*Message, error) {
    
[...]

PR #260, opened 2026-03-14
Don't buffer files in memory at send time

PaulSonOfLars:
# What

Addresses #259 

Previously, we would pipe files while creating the multipart reader.
This changed in #242, where i opted to avoid double-marshalling fields - but introduced a regression where files were double-buffered in memory in an attempt to maintain the fix from #231.

This PR introduces regression tests to cover those edge cases, and then updates the requests code to correctly use pipes and allow retries without double-buffering files.

# Impact

### DIFF
--- a/ext/handler_mapping_test.go
+++ b/ext/handler_mapping_test.go
@@ -62,11 +62,16 @@ func Test_handlerMappings_getGroupsConcurrentSafe(t *testing.T) {
 }
 
 func checkList(t *testing.T, name string, got []Handler, expected ...Handler) {
+	t.Helper()
+
 	if len(got) != len(expected) {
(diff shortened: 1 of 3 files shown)


danielwaichcesl:
Hi @PaulSonOfLars,

Tested PR #260 with a 31 MB file upload - works perfectly! 🎉

Results
Memory usage:

Before: ~66 MB peak (buffered entire file)
After PR #260: ~8.8 MB peak
87% memory reduction ✅
Test details:

31 MB
Upload: successful in ~17s
No memory leaks
GC reclaims all memory correctly
Impact
This solves our issue with large file uploads in containerized environments. The streaming implementation works great - files are no longer buffered in memory.

Thank you for the quick fix! 👏

PR #265, opened 2026-05-12
add original text helpers

d0ctr:
# What

Given that there are already helpers to agnostically get message's text and entities, it makes sense to utilize them for formatting.

# Impact

- One less `if`-statement in user's code!


PaulSonOfLars (approved):

PaulSonOfLars:
Great idea, thank you!

PR #274, opened 2026-07-04
fix(formatting): resolve trailing whitespace reversal bug in splitEdgeWhitespace

Awesome-RJ:
## Description
This PR fixes a bug in the `splitEdgeWhitespace` helper function where trailing whitespace characters (e.g. `" \t"`) were written in reverse order (`"\t "`). Because of this reversal, `strings.TrimSuffix` failed to match and strip the trailing whitespace, leading to whitespace duplication in the final formatted outputs.

### Changes
- Rewrote `splitEdgeWhitespace` in `formatting.go` using index pointers/rune slicing directly.
- Avoided `strings.Builder` allocations and substring searches.
- Created `formatting_test.go` to test whitespace edge cases (mixed whitespace, pre blocks, leading/trailing space).

### DIFF
--- a/formatting_test.go
+++ b/formatting_test.go
@@ -1,0 +1,75 @@
+package gotgbot
+
+import (
+	"testing"
+)
+
+func TestSplitEdgeWhitespace(t *testing.T) {
+	tests := []struct {
+		name    string
+		text    string
+		entType string
+		wantPre string
+		wantCnt string
+		wantPos string
+	}{
+		{
+			name:    "no whitespace",
+			text:    "hello",
+			entType: "bold",
+			wantPre: "",
+			wantCnt: "hello",
+			wantPos: "",
+		},
+		{
+			name:    "leading and trailing",
+			text:    "  hello  ",
+			entType: "bold",
+			wantPre: "  ",
+			wantCnt: "hello",
+			wantPos: "  ",
+		},
+		{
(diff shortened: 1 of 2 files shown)


PaulSonOfLars (approved):
good spot, thank you!

PaulSonOfLars:
@Awesome-RJ Looks like tests aren't passing - are you sure you've defined the expected behaviour correctly?

formatting_test.go
```diff
@@ -0,0 +1,75 @@
...
+			wantCnt: "",
+			wantPos: "",
+		},
+		{
+			name:    "newlines in pre",
+			text:    "\n hello \n",
+			entType: "pre",
+			wantPre: " ",
+			wantCnt: "\n hello \n",
+			wantPos: "",
```
PaulSonOfLars: Are we sure this is the right behaviour? Why would `pre` be an empty space and not `pos`?

Awesome-RJ: Good catch! You're completely right. The code behaves correctly, but I had defined the expected behavior in my unit test incorrectly (asserting wantPre: " " instead of keeping it in wantPos). I've corrected the test case.

PaulSonOfLars:
I've got a small edge case i remember having to deal with which i want to make sure is correctly covered by this - will need to test on my side, but otherwise LGTM :)

PR #276, opened 2026-07-04
sec(request): wrap errors with sanitizedError to prevent token leaks

Awesome-RJ:
## Description
This PR improves token sanitization in `sanitizeError`. The previous implementation modified `url.Error` in-place, which would bypass redaction if the network error was structured/wrapped differently, risking token leaks in logs.

### Changes
- Introduced a custom `sanitizedError` wrapper type.
- Implemented `Error() string` to redact the bot token using `strings.ReplaceAll` for all wrapped errors.
- Implemented `Unwrap() error` so that callers can still use `errors.As` and `errors.Is` to check the underlying error types (e.g. network timeouts).
- Added `TestSanitizeError` unit tests to verify unwrapping and token redaction.

### DIFF
--- a/request.go
+++ b/request.go
@@ -149,7 +149,7 @@ func (bot *BaseBotClient) RequestWithContext(parentCtx context.Context, token st
 
 	resp, err := bot.Client.Do(req)
 	if err != nil {
-		return nil, fmt.Errorf("failed to execute POST request to %s: %w", method, sanitizeError(token, err))
+		return nil, fmt.Errorf("failed to execute POST request to %s: %w", method, &sanitizedError{err: err, token: token})
 	}
 	defer resp.Body.Close()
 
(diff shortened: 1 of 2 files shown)


request.go
```diff
@@ -253,15 +253,25 @@ func buildMultipart(ctx context.Context, params map[string]any) (io.Reader, stri
...
-	if errors.As(err, &urlErr) {
-		urlErr.URL = strings.ReplaceAll(urlErr.URL, token, "<TOKEN>")
-		return urlErr
+	if err == nil {
+		return nil
 	}
-
-	return err
+	return &sanitizedError{err: err, token: token}
 }
```
PaulSonOfLars: We only call sanitise error when handling an error, making the nil check unnecessary - we could just inline this into a struct declaration

PaulSonOfLars (changes requested):
Small comment for readability, LGTM otherwise

request.go
Awesome-RJ: Good point! Since we only call it when `err != nil`, I have inlined the `&sanitizedError` creation directly and removed the helper function entirely.

PaulSonOfLars:
@Awesome-RJ do you intend to fix these PRs?

PR #282, opened 2026-08-27
Patch 8

miscivetprojects-creator:
Thanks for pointing [this ](https://github.com/PaulSonOfLars/gotgbot/pull/278#discussion_r3832307394) out. I double-checked the existing webhook behavior and adjusted the change so webhook and polling registrations are handled explicitly rather than relying on the URL path or polling context.

This preserves the existing behavior for normal webhook paths and also preserves support for a webhook registered at the root path (`""`). I added regression coverage for both the root webhook path and an existing non-root webhook path, as well as ensuring polling bots don't claim a webhook URL.

`go test ./...` passes successfully.


ext/botmapping.go
```diff
@@ -78,8 +78,12 @@ func (m *botMapping) addBot(b *gotgbot.Bot, urlPath string, webhookSecret string
...
 	}
 
-	if _, ok := m.urlMapping[urlPath]; urlPath != "" && ok {
-		return nil, ErrBotUrlPathAlreadyExists
+	// Only webhook bots use urlMapping. This allows an empty urlPath
+	// to be used for a webhook registered at the root path.
+	if ctxClose == nil {
+		if _, ok := m.urlMapping[urlPath]; ok {
+			return nil, ErrBotUrlPathAlreadyExists
+		}
```
PaulSonOfLars, line 86: This seems to make the assumption that ctxClose is always empty for polling bots, but polling bots can also receive ctxClose functions (see `addPollingBot`).

This doesnt seem correct, am i missing something?

ext/botmapping.go
```diff
@@ -230,4 +239,4 @@ func (b *botData) shouldStopUpdates() bool {
 		// otherwise, continue as usual
 		return false
 	}
-}
+}
```
PaulSonOfLars, line 242: please don't strip newlines from files

gen_consts.go
```diff
@@ -114,6 +115,9 @@ func (u Update) GetType() string {
 	case u.Subscription != nil:
 		return UpdateTypeSubscription
 
+	case u.StoppedMessageGeneration != nil:
+		return UpdateTypeStoppedMessageGeneration
+
```
PaulSonOfLars, line 120: Why are these changes included? This doesn't seem relevant to the PR.

spec_commit
```diff
@@ -1 +1 @@
-88d8086c136cd7c6ba6f4ac80fcd12ee33903a4c
\ No newline at end of file
+d59462bb813375e5fd53e3969d59400969148bf7
```
PaulSonOfLars, line 1: The reason the generation is pinned to a commit is so that we can make changes to the generation logic, without being forced to update to the latest version

ext/botmapping.go
```diff
@@ -78,8 +78,12 @@ func (m *botMapping) addBot(b *gotgbot.Bot, urlPath string, webhookSecret string
 		return nil, ErrBotAlreadyExists
 	}
 
-	if _, ok := m.urlMapping[urlPath]; urlPath != "" && ok {
-		return nil, ErrBotUrlPathAlreadyExists
+	// Only webhook bots use urlMapping. This allows an empty urlPath
+	// to be used for a webhook registered at the root path.
+	if ctxClose == nil {
```
PaulSonOfLars, line 83: Is this comment intended to discuss urlMapping, but the if statement acts on ctxClose?

PaulSonOfLars (changes requested):
Hi, thanks for splitting! Could you please use descriptive PR titles?

These changes aren't clearly defined, and it's not clear to me how we'd validate that they aren't backwards breaking. 
Additionally, please dont make unrelated changes to PRs - it makes reviews significantly harder 😕

PR #281, opened 2026-08-27
Fix potential concurrency issue when adding new handlers

miscivetprojects-creator:
Moved the handler slice-copy changes into a separate PR as requested, keeping them isolated from the other changes for easier review.

Related PR: https://github.com/PaulSonOfLars/gotgbot/pull/278/changes/BASE..2f2f7f10ee3238f838a16f8436dd15def6abfa9d#r3866386878

PR #285, opened 2026-09-16
Fix potential concurrency issue when adding new handlers

miscivetprojects-creator:
### What
Fixes a potential concurrency race condition in `handlerMapping.add()` when new handlers are added while updates are being processed.

### Impact
**Are your changes backwards compatible?** Y
**Have you included documentation, or updated existing documentation?** N/A (Internal concurrency fix, no public API changes)
**Do errors and log messages provide enough context?** N/A

### Details
Previously, `m.add()` modified the handler list with:
```go
m.handlers[group] = append(currHandlers, h)


PaulSonOfLars:
Context - duplicate PR of #281 (which was against a non-default branch), which was split out from #278 (which had too many different things happening at once). 

This PR aims to fix an edge case where a possible race condition could happen when adding handlers.


@miscivetprojects-creator  Please make an effort to write up clear PR descriptions - I provide a template for a reason. A PR is not only there to communicate your intent to the PR reviewer, but also to anyone debugging issues in the future. Additionally, the PR description gets used as the commit message on squash-merge, meaning it becomes a permanent part of the git history. You should aim to describe both the issue and how it fixes it.

The easiest way to get a PR merged are for it to be small and well defined. Your first PR was huge and contained multiple changes. Your second PR was opened with no tests whatsoever, against a non-default branch. Now you're opening it with no description, and I'm still the one that's having to add tests. 
The more work you give your reviewers, the harder you make it to get things merged :)

miscivetprojects-creator:
```markdown
Thanks for adding the tests and the `-race` checks!
I apologize for the missing description earlier. I've updated the PR description above using the repository template to clearly explain the race condition and how allocating a fresh slice in `add()` fixes it.

